PR #7, opened 2011-06-21
Added many sample SVG files

cramhead:
HI Matt,

I've never done a pull request and am pretty new to git so hopefully it's all good.
I've integrated many SVG files into the sample that you've provided. The samples are flags so they shouldn't have any gradients; which I believe is a problem now. I'm hope this helps give ammunition for testing and dealing with real world SVG files that are relatively simple. Note: many of the images cause crashes in the SVGKit code.

Cheers,
Marc


iksnae:
none of the SVGs i create work unless they contain only a circle.. what program should i be using to generate compatible SVG files?

cramhead:
Sorry, I didn't realize the circle restriction. In looking that the metadata
in the files it seems many of the images were create with inkscape

Marc

iksnae:
Than you marc.. I was trying Inkscape. It creates several different SVG formats, but none of them worked. After looking at the SVG files in a text editor, I can see xml structure is VERY different from the original.

adamgit:
@cramhead I'm reviewing the old Pull Requests now.

I want to overwrite the main repository with the reklis fork - there are over 1,000 new lines, and 500 changed, since the last update to master!

Unfortunately, I cannot view your Pull Request, it crashes GitHub's webpage. If you're still using SVGKit, could you try re-applying your changes to the reklis fork, and see if it works OK?

Off the top of my head, the main change is that the "samples" are now embedded in a standalone xcodeproj called "SVGPad" (which currently is targetted for iPad only - this is our main testing platform at the moment).

Thanks!

cramhead:
Hi Adam,

Sorry, after a few months I deleted the pull request.
All it included was a new set of sample svg images. They were flags, i.e. simple geometry and no transparency. I could recreate it if it would be helpful. That said, in recreating it I could do so with your new master if once committed.
Let me know.

Marc

adamgit:
The reklis fork has increased SVG support by a large factor - we now
support curves, shapes, polys, some text, some CSS/styles, some
transforms.

So ... yes, it would be great to have some extra tests.

Also, if you have any flags that contain "unusual" features, it would
be great to get them into our test suite too, so we can fix and test.

(I'm guessing there are flags that import bitmaps, or that use
transparency, or use rotation transforms - some/all of those may fail
right now, but since we're trying to fix those features ...)

On 7 May 2012 06:05, Marc
reply@reply.github.com
wrote:

> Hi Adam,
> 
> Sorry, after a few months I deleted the pull request.
> All it included was a new set of sample svg images. They were flags, i.e. simple geometry and no transparency. I could recreate it if it would be helpful. That said, in recreating it I could do so with your new master if once committed.
> Let me know.
> 
> Marc
>

PR #12, opened 2011-10-14
Rendering and Mutability

reklis:
It may not be the cleanest, but it works very well.  I've tested editing SVGs with Illustrator and Inkscape with great results.  It still is not the complete SVG spec, quad braziers are still missing.  Fortunately those appear to be rather atypical.

I've also added mutability and pattern fill so that specifically path layers can be mutated before they are rendered to the layer cache.


adamgit:
PLEASE NOTE: I am intending to CLOSE this pull request WITHOUT MERGING it. This repository is just too far behind - it would take forever to review this many pull requests!

Instead, now that I have commit access to the main repository, I'll replace it with reklis' fork, verbatim.

If you have code that has been merged to the reklis fork already - you don't need to do anything.

Otherwise, please rebase (git rebase?) on reklis, and then you'll be able to easily re-submit your Pull Request once I've updated the main repository.

PR #39, opened 2013-01-18
2013base

mtrubnikov:
Some code refactored; 
Text implementation fixed;
Group elements and transforms done the right way (I hope so);
Matrix transforms parsing fixed;
Added OUTLINE_SHAPES option for debugging purposes - it draws bounding boxes for all elements.

starwars-computer-specialist
May the force be with you. ;)

### DIFF
--- a/Source/DOM classes/Unported or Partial DOM/SVGElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGElement.m
@@ -508,6 +508,10 @@
 {
 	return [self cascadedValueForStylableProperty:@"fill"];
 }
+// TODO:
+// Property "opacity" should be applied to the whole object, not just for the fill.
+// We have "fill-opacity" just for the case. We should use it here.
+// But [self cascadedFillOpacity] returns us the value of "opacity" prop. That's wrong.
 -(NSString *)cascadedFillOpacity
 {
 	return [self cascadedValueForStylableProperty:@"opacity"];
--- a/Source/DOM classes/Unported or Partial DOM/SVGElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGElement.m
@@ -230,7 +230,9 @@
 				return;
 			}
 			NSString* command = [transformString substringToIndex:loc.location];
-			NSArray* parameterStrings = [[transformString substringFromIndex:loc.location+1] componentsSeparatedByCharactersInSet:[NSCharacterSet characterSetWithCharactersInString:@","]];
+            
+            //we MUST take SPACE character into consideration as a parameters separator
+			NSArray* parameterStrings = [[transformString substringFromIndex:loc.location+1] componentsSeparatedByCharactersInSet:[NSCharacterSet characterSetWithCharactersInString:@", "]];
 			
 			command = [command stringByTrimmingCharactersInSet:[NSCharacterSet characterSetWithCharactersInString:@" "]];
 			
@@ -274,25 +276,24 @@
 				 https://github.com/warpflyght/SVGKit/commit/c1bd9b3d0607635dda14ec03579793fc682763d9
 				 
 				 */
-				NSArray *rotateStrings = [[parameterStrings objectAtIndex:0] componentsSeparatedByCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
-				if( [rotateStrings count] == 1)
+				if( [parameterStrings count] == 1)
 				{
-					CGFloat degrees = [[rotateStrings objectAtIndex:0] floatValue];
+					CGFloat degrees = [[parameterStrings objectAtIndex:0] floatValue];
 					CGFloat radians = degrees * M_PI / 180.0;
 					
 					CGAffineTransform nt = CGAffineTransformMakeRotation(radians);
 					self.transformRelative = CGAffineTransformConcat( self.transformRelative, nt );
 				}
-				else if( [rotateStrings count] == 3)
+				else if( [parameterStrings count] == 3)
 				{
-					CGFloat degrees = [[rotateStrings objectAtIndex:0] floatValue];
+					CGFloat degrees = [[parameterStrings objectAtIndex:0] floatValue];
 					CGFloat radians = degrees * M_PI / 180.0;
-					CGFloat centerX = [[rotateStrings objectAtIndex:1] floatValue];
-					CGFloat centerY = [[rotateStrings objectAtIndex:2] floatValue];
+					CGFloat centerX = [[parameterStrings objectAtIndex:1] floatValue];
+					CGFloat centerY = [[parameterStrings objectAtIndex:2] floatValue];
 					CGAffineTransform nt = CGAffineTransformIdentity;
-					nt = CGAffineTransformConcat( nt, CGAffineTransformMakeTranslation(centerX, centerY) );
+					nt = CGAffineTransformConcat( nt, CGAffineTransformMakeTranslation(-centerX, -centerY) );
 					nt = CGAffineTransformConcat( nt, CGAffineTransformMakeRotation(radians) );
-					nt = CGAffineTransformConcat( nt, CGAffineTransformMakeTranslation(-1.0 * centerX, -1.0 * centerY) );
+					nt = CGAffineTransformConcat( nt, CGAffineTransformMakeTranslation(centerX, centerY) );
 					self.transformRelative = CGAffineTransformConcat( self.transformRelative, nt );
 					} else
 					{
--- a/Source/DOM classes/Unported or Partial DOM/SVGGroupElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGGroupElement.m
@@ -53,34 +53,63 @@
 }
 
 - (void)layoutLayer:(CALayer *)layer {
-	CGRect mainRect = CGRectZero;
-	
-	/** Adam: make a frame thats the UNION of all sublayers frames */
-	for ( CALayer *currentLayer in [layer sublayers] )
-	{
-		CGRect subLayerFrame = currentLayer.frame;
-		mainRect = CGRectUnion(mainRect, subLayerFrame);
-	}
-	
-	layer.frame = mainRect;
-	
-	/** Adam:(dont know why this is here): set each sublayer to have a frame the same size as the parent frame, but with 0 offset.
-	 
-	 Adam: if I understand this correctly, the person who wrote it should have just written:
-	 
-	 "currentLayer.bounds = layer.frame"
-	 
-	 i.e. make every layer have the same size as the parent layer.
-	 
-	 But whoever wrote this didn't document their bad code, so I have no idea if thats correct or not
-	 */
-	for (CALayer *currentLayer in [layer sublayers]) {
-		CGRect frame = currentLayer.frame;
-		frame.origin.x -= mainRect.origin.x;
-		frame.origin.y -= mainRect.origin.y;
+	CGRect frameRect = CGRectZero;
+    CGRect mainRect = CGRectZero;
+    CGRect boundsRect = CGRectZero;
+
+	NSArray *sublayers = [layer sublayers];
+    
+	for (NSUInteger n = 0; n < [sublayers count]; n++) {
+		CALayer *currentLayer = [sublayers objectAtIndex:n];
 		
-		currentLayer.frame = frame;
+		if (n == 0) {
+			frameRect = currentLayer.frame;
+		}
+		else {
+			frameRect = CGRectUnion(frameRect, currentLayer.frame);
+		}
+        mainRect = CGRectUnion(mainRect, currentLayer.frame);
 	}
+    
+    boundsRect = CGRectOffset(frameRect, -frameRect.origin.x, -frameRect.origin.y);
+    
+    for (CALayer *currentLayer in sublayers) {
+        [currentLayer setAffineTransform:CGAffineTransformConcat(currentLayer.affineTransform, CGAffineTransformMakeTranslation(-frameRect.origin.x, -frameRect.origin.y))];
+	}
+	
+	layer.frame = boundsRect;
+	
+#if OUTLINE_SHAPES
+    
+    layer.borderColor = [UIColor redColor].CGColor;
+    layer.borderWidth = 2.0f;
+    
+    NSString* textToDraw = [NSString stringWithFormat:@"%@ (%@): {%.1f, %.1f} {%.1f, %.1f}", self.identifier, [self class], layer.frame.origin.x, layer.frame.origin.y, layer.frame.size.width, layer.frame.size.height];
+    
+    UIFont* fontToDraw = [UIFont fontWithName:@"Helvetica"
+                                         size:10.0f];
+    CGSize sizeOfTextRect = [textToDraw sizeWithFont:fontToDraw];
+    
+    CATextLayer *debugText = [[[CATextLayer alloc] init] autorelease];
+    [debugText setFont:@"Helvetica"];
+    [debugText setFontSize:10.0f];
+    [debugText setFrame:CGRectMake(0, 0, sizeOfTextRect.width, sizeOfTextRect.height)];
+    [debugText setString:textToDraw];
+    [debugText setAlignmentMode:kCAAlignmentLeft];
+    [debugText setForegroundColor:[UIColor redColor].CGColor];
+    [debugText setContentsScale:[[UIScreen mainScreen] scale]];
+    [debugText setShouldRasterize:NO];
+    [layer addSublayer:debugText];
+    
+#endif
+
+    //applying transform relative to centerpoint
+    CGAffineTransform tr1 = layer.affineTransform;
+    tr1 = CGAffineTransformConcat(tr1, CGAffineTransformMakeTranslation(frameRect.size.width/2, frameRect.size.height/2));
+    CGAffineTransform tr2 = CGAffineTransformConcat(tr1, self.transformRelative);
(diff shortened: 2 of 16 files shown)


adamgit:
Thanks. There's some good stuff here, but I'll have to cherrypick it - a few bits won't work.

Good catch on fixes eg using a reserved keyword as a parameter (doh)

But some bits I can't merge, will comment for each separately...

Source/DOM classes/Unported or Partial DOM/SVGTextElement.m
```diff
@@ -61,17 +74,20 @@ - (CALayer *) newLayer
 	NSString* actualFamily = [self cascadedValueForStylableProperty:@"font-family"];
 	
 	CGFloat effectiveFontSize = (actualSize.length > 0) ? [actualSize floatValue] : 12; // I chose 12. I couldn't find an official "default" value in the SVG spec.
+    
+    //WHY THE HELL should we do this?! We have transform to be applied to the whole layer!
```
adamgit, line 78: Good question. Experimenting with the examples in the official SVG Spec (which have PNG renders included, so we can check the "expected" output), it seems there's two problems that this solves:
1. SVG Spec _requires_ that we calculate the bounds for the text based on SVG rules. We have to do that using the transform --- I was surprised at this, it makes implementation much harder than I'd like, but SVG spec's approach to text is ... weird
2. Apple _requires_ that you define the text-size BEFORE you render the layer, because Apple's clipping is applied early in the rendering --- I was suprised by this, seems that Apple did some bad API design, and made it so that CATextLayer doesn't have an "automatic" bounds based on the text contents. This is annoying :(.

...put those two together, and the font-size has to be downscaled BEFORE we scale the resulting layer. Yes, its weird - but turn it off, and then try some of the spec examples, and you'll see Apple apply "incorrect" clipping.

Source/DOM classes/Unported or Partial DOM/SVGTextElement.m
```diff
@@ -61,17 +74,20 @@ - (CALayer *) newLayer
...
+//	CGPoint transformedOrigin = CGPointApplyAffineTransform( CGPointMake(self.x, self.y), [self transformAbsolute]);
 	
 	/** find a valid font reference, or Apple's APIs will break later */
 	/** undocumented Apple bug: CTFontCreateWithName cannot accept nil input*/
 	CTFontRef font;
-	if( actualFamily != nil)
-		font = CTFontCreateWithName( (CFStringRef)actualFamily, effectiveFontSize, NULL);
-	if( font == NULL )
-		font = CTFontCreateWithName( (CFStringRef) @"Verdana", effectiveFontSize, NULL); // Spec says to use "whatever default font-family is normal for your system". On iOS, that's Verdana
+	if( actualFamily == nil) {
```
adamgit, line 87: Apple's C API's can fail unexpectedly without returning errors. We have to attempt the font, and then check its null, and then re-attempt if it failed.

Source/DOM classes/Unported or Partial DOM/SVGShapeElement.m
```diff
@@ -71,6 +81,7 @@ - (void)setPathByCopyingPathFromLocalSpace:(CGPathRef)aPath {
 	
 	if (aPath) {
 		_pathRelative = CGPathCreateCopy(aPath);
+        _layerRect = CGRectIntegral(CGPathGetPathBoundingBox(aPath));
```
adamgit, line 84: why is this "integral" ? Both SVG and Apple work in floating point

Source/DOM classes/Unported or Partial DOM/SVGShapeElement.m
```diff
@@ -79,33 +90,19 @@ - (CALayer *) newLayer
...
-	CGRect transformedPathBB = CGPathGetBoundingBox( pathToPlaceInLayer );
-	
-	/** NB: when we set the _shapeLayer.frame, it has a *side effect* of moving the path itself - so, in order to prevent that,
-	 because Apple didn't provide a BOOL to disable that "feature", we have to pre-shift the path forwards by the amount it
-	 will be shifted backwards */
-	CGPathRef finalPath = CGPathCreateByOffsettingPath( pathToPlaceInLayer, transformedPathBB.origin.x, transformedPathBB.origin.y );
-	
-	/** Can't use this - iOS 5 only! path = CGPathCreateCopyByTransformingPath(path, transformFromSVGUnitsToScreenUnits ); */
+    
+	CGPathRef path = CGPathCreateByOffsettingPath(_pathRelative, _layerRect.origin.x, _layerRect.origin.y);
```
adamgit, line 96: We tried this before and it frequently broke (c.f. notes in original code). Does this line work OK for you with all the samples in current branch? (I _believe_ this line broke because of bugs in Apple's code)

Source/DOM classes/Unported or Partial DOM/SVGRectElement.m
```diff
@@ -77,7 +77,7 @@ - (void)postProcessAttributesAddingErrorsTo:(SVGKParseResult *)parseResult {
 	_ry = [[self getAttribute:@"ry"] floatValue];
 	
 	CGMutablePathRef path = CGPathCreateMutable();
-	CGRect rect = CGRectMake(_x, _y, _width, _height);
+	CGRect rect = CGRectMake(0, 0, _width, _height);
```
adamgit, line 80: how does this work? It seems to be ignoring the X and Y attributes.

Source/DOM classes/Unported or Partial DOM/SVGGroupElement.m
```diff
@@ -53,34 +53,63 @@ - (CALayer *) newLayer
...
 
 - (void)layoutLayer:(CALayer *)layer {
-	CGRect mainRect = CGRectZero;
-	
-	/** Adam: make a frame thats the UNION of all sublayers frames */
-	for ( CALayer *currentLayer in [layer sublayers] )
-	{
-		CGRect subLayerFrame = currentLayer.frame;
-		mainRect = CGRectUnion(mainRect, subLayerFrame);
+	CGRect frameRect = CGRectZero;
```
adamgit, line 56: Has this been tested with all the current samples in Demo-iOS project? I don't know preciesly what the original was intended to do, but it seems to be forcing Apple to expand the bounds of each CALAyer to fit its content (because Apple's API isn't very good to start with)

Source/DOM classes/Unported or Partial DOM/SVGElement.m
```diff
@@ -507,6 +508,10 @@ -(NSString *)cascadedFill
 {
 	return [self cascadedValueForStylableProperty:@"fill"];
 }
+// TODO:
+// Property "opacity" should be applied to the whole object, not just for the fill.
+// We have "fill-opacity" just for the case. We should use it here.
+// But [self cascadedFillOpacity] returns us the value of "opacity" prop. That's wrong.
```
adamgit, line 514: Agreed. I'd like to remove these convenience methods, and force all code to call through the "cascadedValueForStylableProperty:" method.

And ... make that method check if the incoming property _CAN_ be cascaded (using the table from SVG Spec)

adamgit:
I've commented on the commit itself, using GitHub's line-by-line.

Some of it is notes to myself - when I get some time, I'll do a careful bit-by-bit merge and test each change.

There's some questions in there too - if you get time, would be great to get your thoughts / help on some of those.

mtrubnikov:
For sure, I'll try to answer each question/comment.

Where, by the way, can I find official examples? 
I tested on files made with Inkscape and CorelDraw.

adamgit:
In the SVG spec itself, ( http://www.w3.org/TR/SVG/ ) they often embed
inline SVG's with PNG files. Each time they do, there's a link either
immediately below - or sometimes a couple of paragraphs away - to "view
this example as SVG (if your brwoser supports it)".

The links resolve to plain SVG that you can save directly.

Also, every time a user reports a "broken" SVG, we ask them to create a
simple example that we can add to the "Demo-iOS" project (almost all of the
files there are ones that used to be broken / fail in SVGKit. A couple of
them still don't work correctly)

Source/DOM classes/Unported or Partial DOM/SVGTextElement.m
mtrubnikov, line 78: Well, OK, but the problem here is that text could not only be proportionally scaled, but all the valid SVG transforms can be applied, like rotation and skew.
I agree, that my implementation is wrong concerning SVG spec bounds calculation... But it lets us RENDER all transforms correctly, breaking layer bounds.

If you create a simple svg with transformed text, you'll easily understand what I mean.
Well, for example (sorry, I don't know, how to share files here):

```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!-- Created with Inkscape (http://www.inkscape.org/) -->

<svg
   xmlns:dc="http://purl.org/dc/elements/1.1/"
   xmlns:cc="http://creativecommons.org/ns#"
   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
   xmlns:svg="http://www.w3.org/2000/svg"
   xmlns="http://www.w3.org/2000/svg"
   version="1.1"
   width="744.09448"
   height="1052.3622"
   id="svg3018">
  <g
     id="layer1">
    <text
       x="14.464138"
       y="80.399139"
       transform="matrix(1.4614298,0.61852241,0.50594217,0.8983918,0,0)"
       id="text3030"
       xml:space="preserve"
       style="font-size:21.16327095px;font-style:normal;font-weight:normal;line-height:125%;letter-spacing:0px;word-spacing:0px;fill:#000000;fill-opacity:1;stroke:none;font-family:Verdana"><tspan
         x="14.464138"
         y="80.399139"
         id="tspan3032">ROTATED AND SKEWED TEXT</tspan></text>
  </g>
</svg>
```

Source/DOM classes/Unported or Partial DOM/SVGTextElement.m
mtrubnikov, line 87: Agreed, my mistake. Just didn't get the point here.

Source/DOM classes/Unported or Partial DOM/SVGGroupElement.m
mtrubnikov, line 56: I've tested this with SOME of the current samples.
The difference is that in my code layer bounds do not include {0,0} point, it just perfectly fits the content, helping us to perform transforms correctly.

You can clearly see the difference with the introduced debug option OUTLINE_SHAPES enabled.

Source/DOM classes/Unported or Partial DOM/SVGRectElement.m
mtrubnikov, line 80: X and Y attributes are taken into account in the parent class (SVGShapeElement, line 194). That was made to have correct layer bounds with transforms. Also, can be seen with OUTLINE_SHAPES enabled.

Source/DOM classes/Unported or Partial DOM/SVGShapeElement.m
mtrubnikov, line 84: No need for it to be "integral". Nevertheless, we need to have _layerRect correctly set.

Source/DOM classes/Unported or Partial DOM/SVGShapeElement.m
mtrubnikov, line 96: I need to check this with all the samples, tested with some of them.

mtrubnikov:
Oh, by the way, do you have any performance issues when zooming in and out? 
With my big svg files with lots of text... It really drives me crazy.

Source/DOM classes/Unported or Partial DOM/SVGGroupElement.m
adamgit, line 56: A lot of Apple's API requires 0,0 to be included, otherwise all the co-ordinate systems are broken.

Transforms are currently correct in all the samples. I'll try merging OUTLINE-SHAPES (looks very useful!), but also: do you have an example SVG that renders incorrectly with 0,0 included? would be good to add as a test case

Source/DOM classes/Unported or Partial DOM/SVGTextElement.m
adamgit, line 78: Yeah, I see what you mean.

Perhaps we need to pull your last few lines in this method (which do the transforms) up to this point, and apply them to the results of Apple's CoreText methods from the 2013base code ... and work out the actual box needed to send to CATextLayer / etc?

adamgit:
With SVGKFastImageView - no, that's why I made it :). I haven't tried it
with text, but it works well with VERY complicated shape/path data.

If FastImageView is slow, it suggests that CATextLayer is broken and
useless (thanks to Apple) and that we'll have to re-implement it using
CoreText :(.

Source/DOM classes/Unported or Partial DOM/SVGGroupElement.m
mtrubnikov, line 56: Well, actually, as it seems to me, all the coordinate systems with groups are broken now. Just because they are not transformed.
Also, instead of just transforming CALayer for group element, we need some magic like:

```
for (CALayer *currentLayer in [layer sublayers]) {
   CGRect frame = currentLayer.frame;
   frame.origin.x -= mainRect.origin.x;
   frame.origin.y -= mainRect.origin.y;
   currentLayer.frame = frame;
}
```

They have a good explanation here: http://www.w3.org/TR/SVG/coords.html#EstablishingANewUserSpace
(pictures are really useful!)

It seems to render correctly, though (at least, until there's no transformed text). I will try to draw an example.

Source/DOM classes/Unported or Partial DOM/SVGTextElement.m
mtrubnikov, line 78: We should try.
As far as I see, according to spec, we have to get bounds look somehow like the red box.
With my code, we get the green one.

drawing

adamgit:
Just a quick note to say: I'm swamped with work right now :(, haven't had time to finish merging this. But will get back to it ASAP when I can.

Source/DOM classes/Unported or Partial DOM/SVGShapeElement.h
```diff
@@ -24,6 +24,10 @@
 	CGRect _layerRect;
 }
 
+@property (nonatomic, readonly) CGFloat x;
```
adamgit, line 27: Accoridng to SVG spec, there's no X nor Y in SVGShapeElement - they only exist in SVGRectElement

(some of these SVG*Element classes are badly wrong in SVGKit right now, but I'm moving them all towards being spec-compliant)

Source/DOM classes/Unported or Partial DOM/SVGTextElement.m
adamgit, line 78: So ... I've spent a long time looking at this.

The short answer is: your original comment "why do we do this?" is because you've misunderstood how the SVG library works.

The library - from day 0 - was built to do this:
1. work out the position and transofrmation of everything
2. position everything correctly on screen. If something needs rotating or skewing, do that to the shape itself
3. build CALayers that make things appear in those places. NEVER SET THE TRANSFORM FOR ANY LAYER

Advantages:

a. Users can use 3D transforms, or custom Affine transforms, and it all works
b. Render speed and hit-test speed is - theoretically - much faster because the transforms only have to be calculated once

Disadvantages;

c. If you take a layer that's invisible (e.g. <G> elements are defined as invisible in SVG), and make it visible - it will appear to be in the wrong place.
d. We need the "layoutLayer" method to be called on EVERYTHING after making the layers - because we don't know the correct size of ANY layer until we've rendered ALL layers

adamgit, line 78: I don't know that the original version is "best" - but I've looked at the amount of code needed to change it (a lot!) and ... it would seem to make the library slower, so ... I'm going to undo my changes and leave it working the way it was previously.

adamgit:
FINALLY merged ... great work, @mtrubnikov !

Because GitHub refused to auto-merge, I had to split it up into small pieces. There is ONE PIECE I haven't done which should work fine - I'd prefer you to submit it as a separate commit (so it shows up under your username) -- the two UIColor-Expanded files.

For the rest:

"Some code refactored;" -- I had to skip the changes to layoutLayer because they triggered some crashes elsewhere. I don't know why, I'll have to revist this later :(

"Text implementation fixed;" -- in the end, I used your approach of setting .affineTransform - but to make it work with ALL our text examples I had to do it differently (c.f. the code comments for detailed explanation of WTF I did and why)

"Group elements and transforms done the right way (I hope so);" -- as discussed, there's too many problems with Apple's .affineTransform code for us to use it throughout SVGKit. However, inspired by your fixes, I re-wrote the whole of transforms to be MUCH easier to read, and to be MUCH closer to SVG Spec compliance (c.f. the new protocol "SVGTransformable" that's part of the SVG Spec)

"Matrix transforms parsing fixed;" -- merged fine, but it had some problems with complex transforms, so I had to tweak it a little. Seems to work.

"Added OUTLINE_SHAPES option for debugging purposes - it draws bounding boxes for all elements." -- this is a great idea, but it can be done in user-code (as a runtime thing). I've moved your code into the Demo-iOS project, AND I've merged it with some code I wrote for hit-testing. Now whenever you tap anywehre on screen, it outlines the smallest containing object.

mtrubnikov:
Sorry, I had no time even to answer you here recently.
Thanks!
I'll try to commit the two above mentioned files as soon as possible.

adamgit:
No worries - if you don't have time, I can commit them. But I prefer to keep the commit history accurate, if possible.

PR #55, opened 2013-02-21
keyvalueobserver bug fix. refer to  issue 42

tsaizhenling:
it is possible to initialize SVGKFastImageView with a null SVGKImage.
when another instance of SVGKImage is assigned. it is not added as observer, but is removed as observer when SVGKFastImageView is deallocated.


adamgit:
Nice catch, thanks.

PR #57, opened 2013-02-27
memory leaks

tsaizhenling:
fixed memory leaks. SVGKit is till leaking. disable global cache reveals more leaks


Source/Parsers/SVGKParser.m
```diff
@@ -83,7 +83,10 @@ - (void)dealloc {
 	self.source = nil;
 	[_storedChars release];
 	[_stackOfParserExtensions release];
+    [_parentOfCurrentNode release];
```
adamgit: This is never retained - should not be released?

Source/SVGKImage.h
```diff
@@ -46,7 +46,7 @@
 #import "SVGKSource.h"
 #import "SVGKParseResult.h"
 
-#define ENABLE_GLOBAL_IMAGE_CACHE_FOR_SVGKIMAGE_IMAGE_NAMED 1 // if ENABLED, then ALL instances created with imageNamed: are shared, and are NEVER RELEASED
```
adamgit, line 49: I understand you might not want to use this feature - that's why it's a #define (set it to 0 to compile it out).

But, for the main project, we need this turned ON by default, sorry. Can't merge this :(

adamgit:
Good work, it's 99% great - but: I think you have 1 too many releases, and we have to leave the #define ON by default.

Please check those two items (I've commented in-line) and update, then I can merge this

Thanks!

Source/SVGKImage.h
adamgit, line 49: Sorry, to be clear: it's a good idea to turn this off for testing - but I can't merge it, and Github won't let me "pick and choose" :(

Source/Parsers/SVGKParser.m
tsaizhenling: yes you are right

Source/SVGKImage.h
tsaizhenling, line 49: done

tsaizhenling:
Screen Shot 2013-03-02 at 12 15 25 AM
 left with these memory leaks. I can't quite figure out how to fix these :(

Issue #61, opened 2013-03-11
Load SVG content from memory


rakeshta:
i have an app that generates SVG content in memory. I would like to render this content directly from memory.

Right now my only option is to write the content to a temporary file and then load it again.
This is costly for performance.

rakeshta:
I tried implementing my own `SVGKSource` but that results in an `EXEC_BAD_ACCESS` crash when the `SVGKImageView` is released.

The crash happens on the first line in `[SVGKImage dealloc]` that tries to access `self.DOMTree`

rakeshta:
Here is my implementation of `SVGKSource`

```
@interface VPSVKSourceNSStringReader : NSObject <SVGKSourceReader> {
    NSData *_data;
    NSUInteger _carat;
}

@end

@implementation VPSVKSourceNSStringReader

- (id)initWithString:(NSString *)string {
    if (self = [super init]) {
        _data  = [[string dataUsingEncoding:NSUTF8StringEncoding] retain];
        _carat = 0;
    }
    return self;
}

- (int)readNextChunk:(char *)chunk maxBytes:(int)maxBytes {
    NSRange range = NSMakeRange(_carat, MIN([_data length] - _carat, maxBytes));
    [_data getBytes:chunk range:range];
    _carat += range.length;
    return range.length;
}

- (void)dealloc {
    [_data release];
    [super dealloc];
}

@end


#pragma mark -

@interface VPSVKSourceNSString : SVGKSource {
    NSString *_svgString;
}

@end

@implementation VPSVKSourceNSString

+ (VPSVKSourceNSString *)sourceWithString:(NSString *)svgString {
    return [[[self alloc] initWithString:svgString] autorelease];
}

- (id)initWithString:(NSString *)svgString {
    if (self = [super init]) {
        _svgString = [svgString retain];
    }
    return self;
}

- (void)dealloc {
    [_svgString release];
    [super dealloc];
}


#pragma mark -

- (NSObject<SVGKSourceReader> *)newReader:(NSError**)error {
    return [[[VPSVKSourceNSStringReader alloc] initWithString:_svgString] autorelease];
}

- (void)closeReader:(NSObject<SVGKSourceReader> *)reader {
}

- (int)reader:(NSObject<SVGKSourceReader> *)reader readNextChunk:(char *)chunk maxBytes:(int)maxBytes {
    VPSVKSourceNSStringReader *stringReader = (VPSVKSourceNSStringReader *)reader;
[...]

PR #66, opened 2013-04-02
Switch from SVGKSourceReader to NSInputStream

alejandro-isaza:
Using NSInputStream seems like a better approach than using a custom reader class.


alejandro-isaza:
Fixes #61 as well

PR #80, opened 2013-05-12
Mac Port

MaddTheSane:
I've created a new branch to prevent me from flooding the SVGKit people.

This patch adds Mac support with a Mac OS X framework.


adamgit:
Someone mentioned there were problems with the OS X linking when sharing code with main source project.

Yesterday, I hit similar problems on live projects when we upgraded from Xcode 4.6.0 to 4.6.2 - it seems that 4.6.1 or 4.6.2 changes the linking behaviour (!) for subprojects.

I'll do a review of this before merging, but can you confirm what version(s) of Xcode you're using, and whether the linking is working OK for you at the moment?

(if there's a version-specific problem with Xcode, I'd like to get as much info as possible before I try to find a workaround :))

MaddTheSane:
I am using 4.6.2, and am having no issues linking my image rep bundle to the framework. Of course, I also am ahead of this branch: I didn't want to overwhelm you with my constant commits.

adamgit:
I've merged the other pull requests, I want to start merging this now.

I've built a merge branch, and I've resolved all the conflicts.

HOWEVER the Demo-iOS project is now broken. It is an "impossible" situation - clang won't build because of errors in a file THAT DOES NOT EXIST in the project. It appears to be a (major!) bug in Xcode: it's somehow grabbing a header file from the SVGKit-OSX project and using it in the Demo-iOS project _eventhough_ there is literally no link and no reference between the two projects/files.

Do you have any ideas what's happened / how to fix it? This completely blocks any merging of the branches :(

https://github.com/SVGKit/SVGKit/commits/merge-MaddTheSane-to-1.1.0

adamgit:
The problem:

Something in the "Demi-iOS" project is now tricking Apple into saving all the intermediary build data into the project folder!

This is very wrong, but I can't find a way to turn it off. It means you can compile the project ONCE and then never again (because the "build" sub-folder confuses Xcode, and breaks everything)

MaddTheSane:
At least for me, the demo-iOS project is trying to include my "AppKit Additions" headers by accident. I found out that this is because the Demo-iOS project recursively includes everything in Source, including the AppKit additions.

adamgit:
Ah, yes, I see. IIRC that was to work around a bug in Xcode where it kept removing files from the build even though they were in the project (and you had to manually "remove/re-add" them each time Xcode did this). Telling Xcode to "just compile everything in Source/" fixed it.

adamgit:
There were a few odd and wrong things in the Header Search Paths.

But ... cleaning that out, iOS can't build any more, because Xcode's longstanding bugs in finding headers for sub-projects.

(Xcode won't let you add headers, and it won't look for headers in the standards locations, nor in its own build folders, nor in the project. It changes from version to version too :( )

If I can't find a fix for this, we might have to split "Source/" into 3: "Source/Core", "Source/iOS-Additions", "Source/OSX-Additions"

NB: this is what the original SVGkit used to do. I don't know for sure, but I suspect it was to workaround the Xcode compilation and linking bugs.

adamgit:
Got it. Adding a User Header Search Path of "$(TARGET_BUILD_DIR)/usr/local/include/" (NB: quotes are required, Xcode doesn't automatically support folders with spaces!) fixes it. /usr/local/include is the folder structure that (supposedly) is standards correct and is exported to by the universal (iPhone + simulator) build-script, but Apple seems to prefer "/include" by default. I might change that later.

Now onto your commits ... I've got some crashing bugs (looks like over-release of memory or similar), probably something when I merged with latest 1.1.0 code.

adamgit:
Fixed the mem crash. This now seems to me correct with 1.1.0 :

https://github.com/SVGKit/SVGKit/commits/merge-MaddTheSane-to-1.1.0

But there's no "Demo-OSX" project, so I cannot check _any_ of the changes on the OS X code, cannot see if they are working OK. Please can you try that branch and confirm that the OS X side is still working?

NB: going forward, we urgently need a "Demo-OSX" project that loads each of the SVGs in the "Demo-Samples/SVG/" folder and sub-folders. Otherwise we have no way of checking people's commits.

adamgit:
Going through the diffs, it appears you've removed lots of "autorelease" that were correct originally. Is there a specific reason for this?

adamgit:
Also: you've replaced a lot of "CGFloat = ... [something floatValue]" with " SVGKCGFloatValue]" - but a lot have been left untouched.

(EDIT: or maybe this is due to something going wrong with the merge?)

Is this something that should be changed everywhere? e.g. SVGElement is a messy mix of the two approaches right now.

(Without a demo OS X project to test against, I don't want to make that change)

adamgit:
Also: when you convert lots of 2-line:

"object = ... 
return object;"

into 1-liners ... I thought the 1-liner prevents Xcode's debugger from telling you the value before it returns - or is there  a feature in Xcode that makes this work even with 1-liners?

MaddTheSane:
For the autorelease removals, I did it because I didn't want the iOS autorelease pools to be too big.

As for the return values, I think there is a way to see what the return value, or it shows the return value when you step out of the function. If this has changed, I can change those values back.

For the SVGKCGFloat, I tried to get all values that were CGFloats. I purposefully skipped over the SVGRect creation function because SVGRects right now are just floats.

MaddTheSane:
As for writing an OS X demo project, I'll get around to that as soon as I can.

adamgit:
OK, so ...
- SVGKCGFloat: I'll try this tomorrow, should be quick to do.
- iOS autorelease pools: so far, this is not an issue - I've loaded
  massive SVG files on iOS. But I've got a feeling we're not draining the
  global release pool after parsing - which is stupid (if it's the case), we
  should definitely wrap an autoreleasepool around the entire "parse" stage,
  and then drop the lot, that would be a good move I think?
- return 1 liners: if it's easy to back-out, could you back that out for
  now, and if/when someone figures out what to do with Xcode, we'll add it to
  the project README, and re-do the 1liners?

adamgit:
Re: demo - I'd offer to do it myself, but I'm very rusty at OS X coding :(

FYI: we wanted the iOS demo to show "here's how to use the library", but
most people ignore it and use other resources. So... don't worry about the
code quality - just make the simplest OS X app you can, so long as it lets
you view each of the SVG files (the iOS demo uses Apple's default "new app"
template with small modifications).

But the iOS demo has saved us many times when people have made pull
requests that accidentally broke something. Also very useful for testing
new features - let's you check that they work :).

MaddTheSane:
I have created an OS X demo app, which also made me realize that the layered view wasn't working on OS X (I have fixed it).

MaddTheSane:
Before you merge my code (because I have a **lot** of changes), merge other people's, then let me know so I can merge them on my branch and get them working.

adamgit:
@MaddTheSane - what's the current status on this?

iOS 7 will go live soon, along with Xcode 5.

Xcode 5 will _almost certainly_ break the linking process (at least
for iOS - Apple has changed it in every major release of Xcode so
far). I feel this would be a good opportunity to package up a 2.0
release of SVGKit, for both OS X and iOS, merge all the patches, etc.

Thoughts?

MaddTheSane:
I've just checked out the new 1.1.0 code and will merge it shortly. Note that you might want to modify the read-me, as mine is specific to my branch.

MaddTheSane:
There, the merge should be fairly painless against current code in the 1.1.0 branch.

adamgit:
You have 260 pages of diffs.

I cannot merge that. I'm not going to undermine the live projects that are
using SVGKit by blindly committing changes we've not checked.

Please chop it into feature-oriented, labelled commits, that we can check +
merge.

I would love to get your fixes, but if you can't/won't let us see what they
are, I'm afraid we'll just have to re-write them ourselves and commit them
one by one.

adamgit:
I've added a link to your fork to the main README on front page (which now
automatically points to the current development branch).

I've also added a note requesting help with checking/porting/integrating
the changes - hopefully we'll get some volunteers and we can get everything
checked and added.

MaddTheSane:
Unfortunately my computer is having issues so it will be awhile before I can do anything.

MaddTheSane:
This has too many commits for it to be merged, and the 1.1 branch isn't going to be updated with these requests. Closing.

PR #82, opened 2013-05-13
Add Lumberjack Logging Framework

dblandin:


dblandin:
There are a few features missing from this pull request at the moment:
- Allow the client to specify which log level SVGKit should use internally

Something like:

``` objective-c
[SVGKit enableLogging: SVG_LOG_LEVEL_VERBOSE];
```

Right now, the log level is set to **verbose** for DEBUG builds, and **warn** for all other builds.
- Use a [custom context](https://github.com/robbiehanson/CocoaLumberjack/wiki/CustomContext) for SVGLogWarn-style (instead of the default DDLogWarn) syntax.

The ddLogLevel constant is currently defined at compile time. I'm not sure of the best way to make this a dynamic option set by the client app and globally available in SVGKit.

Any thoughts?

adamgit:
Is it possible to include as a binary instead of source code? (I don't know if that's allowed from the license). Including full source can cause problems later if you try to integrate with other projects that also include the full source (I'm not 100% sure why this happens, seems to be something odd in houw Xcode links sub-projects).

Github won't auto-merge this, so I'll have to come back later and merge by hand, but otherwise looks fine.

dblandin:
Looks likes redistribution in binary form is allowed. 

I'll see if I can convert the source code to a universal binary and use that instead.

adamgit:
Assuming this works, of course :). I'm only trying to find whatever is the
easiest, most robust way of shipping the lib. If source-code is the only
way to do that, we should do it that way.

(might be we just have to include the source, but make sure we don't
accidentally export any of the symbols?)

dblandin:
I wasn't able to find a way to include Lumberjack as a binary without losing the ability to adjust the LogLevel (which takes place during compilation).

Right now, the LogLevel is set here:
https://github.com/dblandin/SVGKit/commit/838496a160cd8693bc17325866ecf9f647406779#L0R10

adamgit:
I haven't had time to review recently, sorry - very busy in day job. When I get time soon, I'll review all the outstanding requests and try to get them all merged + tested in a single batch.

(if anyone wants to help, you can take the "1.1.0" branch, merge the change requests LOCALLY, run the Demo app and test that every image is working fine, then contact me and let me know. We work on the basis "get a 2nd person  to check the code before we merge it")

markst:
If I have CocoaLumberjack as part of my project as well as `SVGKit`, then the global variable `ddLogLevel` causes conflicts.

seltzered:
Hey @markst is this occurring on the latest 2.x branch? Are you installing svgkit via cocoapods or something else? 

What version of CocoaLumberjack are you using in your own project? In svgkit 2.x I updated things a few months ago to use CL 2.0 (see https://github.com/SVGKit/SVGKit/pull/251 )

PR #83, opened 2013-05-15
Switch to ARC

### LINKED ISSUE #20: ARC Compatibility
Hi,

I download to try SVGKit and i found have not compatibility with iOS 5 with ARC.

Will be compatible in a near future?

Thanks!
Manu

alejandro-isaza:
I made the necessary changes to switch to automatic reference counting (aka. ARC). It's mostly removing code so I wouldn't say it's a risky change. In the way I fixed a few bugs, besides all the potential memory leaks:
- Some setters where not correctly handling the case where you reassign the same object. For instance `CSSValueList` `setCssText:`. If you do list.cssText = list.cssText the property would have been released creating a dangling reference.
- In `SVGKParser.m` `errorEncounteredSAX` a c string was being casted to an NSString directly probably causing ugly memory problems.

Fixes #20


adamgit:
ARC is not backwards-compatible (we've had a major crash recently caused by ARC), so we need to maintain a clean, non-ARC, master.

I'm looking for a good solution here. Do you think it would work if we do this:
1. I add a new branch to SVGKit/SVGKit: "master-ARC" (which is derived from "master" and always tracks it)
2. Whenever we release a new tagged version - e.g. "1.1.0" - we also release an ARC that's based off that - e.g. "1.1.0-ARC"

...so that way, you can always pick "ARC" or "non-ARC" version of the source to use in your own projects ... AND (I think?) ... when updating ARC version, it's very easy: the latest-ARC-version is always "close to" the master version.

What do you think - would that work?

(Sorry to be a pain, but we've got people who are still using old versions of Xcode, and ARC isn't an option for them.)

alejandro-isaza:
Yeah, that's an option, it should be relatively easy to keep the ARC branch in sync. I can also just keep my ARC branch and update it periodically in case anyone is interested.

adamgit:
I've created an official "master-ARC" branch as noted above. It currently is based off 1.1.0 (NB: "master" currently tracks 1.0.0 - that's a mistake, and will be fixed very soon)

I've merged your arc to make the initial master-ARC. Hopefully we can keep this updated with changes to master as we go forwards

PR #96, opened 2013-07-31
support for radial gradients

tsaizhenling:
masked gradients in svgkfastimageview also works now


tsaizhenling:
closed #95  and #93  cause i was still figuring out how forking works.. sorry:(

adamgit:
So, if I understand this correctly...
1. Apple has a CAGradientLayer class that _does not support_ radial gradients
2. Apple's CG library _does support_ radial gradients
3. You've created a CAGradientLayer subclass that _adds support for_ radial gradients

?

tsaizhenling:
yes that's right

Issue #99, opened 2013-08-07
Doesn't compile in certain iOS versions


buscarini:
SVGRectElement declares a function CGPathAddRoundedRect, which might have been added in “some” iOS beta versions

zeeshanlakhani:
Yeah. `CGPathAddRoundedRect` is throwing errors on build of some of the beta versions.

miwillhite:
+1

IamAliSufyan:
Facing same build errors. Please fix it

adamgit:
Personally, I don't run the iOS betas (too many bugs until you get to the very final one).

If you want to rename that method, go for it - do a global rename in the source, check the Demo still works (it has an image that uses roudnded rects and nothing else, should be a good test). 

Or even better: see if the method can be deleted, and we can use the new Apple one instead. Use a #if check to see which version of iOS it is, and use the Apple one for new iOS, our one for old iOS.

Github will turn it into a pull request for you, I'll give it a quick check, and merge immediately if it seems to work.

PR #105, opened 2013-09-19
Hotfix for CGPathAddRoundedRect compile error on iOS 7.

MartinMReed:
Added iOS version check to use the built-in CGPathAddRoundedRect when compiling for iOS 7. This commit resolves issue #99.

### DIFF
--- a/Source/DOM classes/Unported or Partial DOM/SVGRectElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGRectElement.m
@@ -6,7 +6,9 @@
 
 @interface SVGRectElement ()
 
+#if __IPHONE_OS_VERSION_MAX_ALLOWED < __IPHONE_7_0
 void CGPathAddRoundedRect (CGMutablePathRef path, CGRect rect, CGFloat radiusX, CGFloat radiusY);
+#endif
 
 @end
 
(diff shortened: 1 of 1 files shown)


MaddTheSane:
Is the master branch the one we should be using as a baseline?

adamgit:
Argh, no - sorry, I thought it was against 1.1.0 or I wouldnt have approved
it. My mistake.

MartinMReed:
Why not, it was a hotfix? Master wasn't even usable on iOS7 without it. iOS7 is no longer in beta, which was your original reason for not fixing it.

adamgit:
Sure, and you didnt do anything wrong but i should have spooen to
you. You're possibly making assumptions about the contents of master that
are reasonable but incorrect. I've tried getting evryone to use it the same
way, but too many people dont read instructions, and it became a mess.

I dont blame them. Github has no easy way of saying "development branch is
X, live branch is Y", and trying to use some standard git approaches has
failed because of that, and because different people have incompatible
ideas about what master should point to.

(Whats worked very well on commercial projects, where we all work together
in one team, doesnt work where every contributor is separate)

So ... The plan is to re invent master in a way that will be misinterpreted
by the smallest number of people. Thats non trivial, both working out
"what" and also shuffling branbes to make it true. Until then, we've been
saying its to be avoided, and Ive been pushing people to do all changes
against 1.1.0 branch.

It might make sense to additionally backport something to master, but not
to do it there first.

MartinMReed:
Okay. Under the repo settings you can change the default branch so people will arrive at 1.1.0 instead of master. Having it presented first will probably help funnel new commits into it.

adamgit:
Cool, that would be a big help, thanks

Issue #106, opened 2013-09-25
Assertion failure on iPhone 5s at pixelsPerInchForCurrentDevice


rrbrambley:
iPhone6 is not an acceptable prefix for the value of hw.machine, causing this assertion failure: NSAssert(FALSE, @"Not supported yet: you are using an iPhone that didn't exist when this code was written, we have no idea what the pixel count per inch is!");

adamgit:
Appears fixed, not sure why this is still open, closing...

PR #107, opened 2013-09-25

rrbrambley:
This fixes issue #106.


MartinMReed:
You might want to switch this to the 1.1.0 branch or it might not get accepted. (See https://github.com/SVGKit/SVGKit/pull/105#issuecomment-25656968)

rrbrambley:
Good call, thanks.

PR #110, opened 2013-10-05
Skipping cache purge and log message when cache is empty

MartinMReed:
The warning message can be misleading at first when theres nothing actually purging. Also, shouldn't need to call removeAll when its already empty.


adamgit:
Good catch. Could you please resubmit this against the 1.x branch (that's where all active dev happens)? Github won't let me merge direct onto it :(.

(obviously I could copy/paste your change myself since it's a 1-line diff, but I'd rather preserve the credit!)

MartinMReed:
Resubmitted as pull request #126

PR #111, opened 2013-10-05

rrbrambley:
This fixes issue #106.

### DIFF
--- a/Source/DOM classes/SVG-DOM/SVGLength.m
+++ b/Source/DOM classes/SVG-DOM/SVGLength.m
@@ -141,7 +141,8 @@ static float cachedDevicePixelsPerInch;
 		return 163.0f;
 	
 	if( [platform hasPrefix:@"iPhone4"]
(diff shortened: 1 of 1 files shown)


bdalziel:
I haven't checked, but is iPhone6 the identifier for the 5C and 5S? I don't know either way, but i just submitted this before i noticed yours:

https://github.com/SVGKit/SVGKit/pull/112

bdalziel:
It must be, ignore me - thanks for posting the correct pull request.

cbhl:
@bdalziel: AFAICT, the iPhone 5C still returns a string starting with "iPhone5", but the iPhone 5S returns a string starting with "iPhone6". (Perhaps due to things like having a 64-bit processor?)

adamgit:
Ah, nicely caught!

Can you re-submit this against 1.x branch?

(I've fixed github so it always defaults to this branch now, correctly, as the active commit branch for development)

Sorry - I can't seem to get github to let me re base it automatically :(

PR #116, opened 2013-10-18
Not include unnecessary files in SVGKImage.h so app can contain only it.

rhcad:
1.  Define __IPHONE_7_0 for iOS 6 to avoid implicit declaration warning about CGPathAddRoundedRect.
2.  Not include unnecessary files in SVGKImage.h so app can contain only it.

### DIFF
--- a/Source/DOM classes/Unported or Partial DOM/SVGRectElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGRectElement.m
@@ -6,6 +6,9 @@
 
 @interface SVGRectElement ()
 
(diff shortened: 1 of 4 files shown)


adamgit:
Thanks. Please could you re-submit this against the 1.x branch?

We don't currently use "master" - it is pending a major clean-up (due to historical reasons, and the need to merge 3 years of patches all in one go!)

I think we already have a workaround for the iOS7 issue - but removing all those includes from SVGKImage.h looks like a good idea to me, worth merging with 1.x

PR #117, opened 2013-10-18
Clear SVGKImage cache when app enters background.

florianreinhart:
I have added code to clear the SVGKImage cache when the app enters background or receives a memory warning. This is useful because an app that consumes less memory, is less likely to be purged from memory while in background.

You could also use NSCache to cache images instead of NSMutableDictionary, which will automatically remove cached objects.


adamgit:
Nice idea.

Could you please re-submit this against the "1.x" branch, instead of 1.1.0? I tried to shift it over in GitHub, but the GUI doesn't seem to let me.

(then I can merge it immediately)

PR #120, opened 2013-11-01
Add support for external CSS stylesheets and marginally improved CSS parsing.

dgileadi:
This pull request supports external CSS stylesheets by relative path. CSS rule parsing is also improved to support element and ID selectors in addition to class selectors, as well as combinations of selectors.

Note that if an external stylesheet isn't found at a relative path to the original SVG file it is then searched for in the bundle and then in the user's document directory. This allows NSString-sourced SVG to specify an external stylesheet, and also allows apps to write custom stylesheets at runtime which can be loaded by the parser.

SVGKImage is also improved to better support initWithCoder, allowing good behavior when SVGKImages are placed in NIB/XIB files.


adamgit:
This looks like some great improvements, I'd like to merge them.

I'm sorry to be a pain, but - can you refactor this slightly so that you aren't modifying ANY of the files/classes/headers in "Core DOM"?

Specifically, those header files are _defined_ by the XML-DOM spec! The names, the types, everything is part of the Spec.

(if one of the files is incorrect, different from the spec - it needs fixing. But I believe they are currently all correct?)

I don't know enough about the CSS selection stuff to say if/how you could avoid editing the CoreDOM headers and classes. If you _absolutely have to_ edit them, the way around it is to create Objective C Categories for the affected classes, and we can put them in a different folder, so its clear they aren't part of the spec.

However, so far ... I've been working very hard to avoid even that. I've been trying to push all "not in the DOM spec" code into standalone classes (which, ideally, we'll put in a different folder).

dgileadi:
Yes, I think I can make it work without modifying the Core DOM files. I'm a bit busy with other things so it may take a few days.

And thanks for your part in such a cool library!

dgileadi:
Replaced by new pull request. Sorry if that's bad etiquette; I'm a bit new to github.

adamgit:
Yes, that's fine, thanks.

PR #122, opened 2013-11-09
Code clean-up

MaddTheSane:
Some minor code clean-up.


adamgit:
Github won't merge it, but won't say why :(.

The only problem I see is maybe the autoboxing - if that overlaps with other changes, maybe it's confusing git-merge. It's also incompatible with the rest of SVGKit - we left it out so it builds on 4.x.

Perhaps we can now say "Xcode 4.5 is the new minimum supported for SVGKit" ? I know that fairly recently we still had people using Xcode 4.2, 4.3 - but maybe we can now move on?

(I've no idea - I don't even know how to find out how many people are using which version :( )

MaddTheSane:
@adamgit One way to find out which version of Xcode people are using is to ask people.

Also, I've merged the most recent version of the 1.x branch to my clean-up branch.

adamgit:
Do you have contact info for people using SVGKit? I don't think we've ever
collected email addresses, and I can't find anything in GitHub for
messaging watchers/forkers :(

MaddTheSane:
I seem to have forgotten this is here. Closing.

PR #124, opened 2013-11-19
Add support for external CSS stylesheets and marginally improved CSS parsing, "Leave Core DOM Alone" edition

dgileadi:
This pull request supports loading external CSS stylesheets by relative path. CSS rule parsing is also improved to support element and ID selectors in addition to class selectors, as well as combinations of selectors.

Note that if an external stylesheet isn't found at a relative path to the original SVG file it is then searched for in the bundle and then in the user's document directory. This allows NSString-sourced SVG to specify an external stylesheet, and also allows apps to write custom stylesheets at runtime which can be loaded by the parser.

SVGKImage is also improved to better support initWithCoder, allowing good behavior when SVGKImages are placed in NIB/XIB files.

This pull request does make one change to a file in Core DOM, CSSStyleSheet.m, to fix a bug where the selector text has its first character chopped off. The old behavior makes sense if you expect your selector to always be a class selector, but completely breaks anything else.


adamgit:
This seems to have broken SVGKLayeredImageView.

If you run the demo included with SVGkit, and scroll down to the bottom, there's a test "nil-demo-layered-imageview". That is coming up as a blank image - it should be displaying a garish, ugly, "YOU SUPPLIED A NIL IMAGE, FOOL!" error image instead.

From the diffs, you haven't removed this feature, but somehow you're bypassing the code that checks for this situation and inserts the dummy / error SVG. Can you check and see if you can get it working again please?

(it's possible someone else has removed it, but I checked this image in a commit only a couple of days ago, and it was working fine - apologies in advance if it was broken by a different recent change!)

dgileadi:
Yes, I'll check it out.

dgileadi:
This was indeed caused by me. Created a new pull request that fixes it.

PR #125, opened 2013-11-19
Fixed the demo issue mentioned in pull #124.

dgileadi:
I'd tend to argue that the demo should be calling initWithSVGKImage with a nil image instead of initWithCoder with a nil coder, but regardless this encouraged me to clean up populateFromImage in both kinds of SVGKImageViews.

### DIFF
--- a/Source/UIKit additions/SVGKFastImageView.m
+++ b/Source/UIKit additions/SVGKFastImageView.m
@@ -51,33 +51,6 @@
     return nil;
 }
 
-- (void)populateFromImage:(SVGKImage*) im
-{
-		internalContextPointerBecauseApplesDemandsIt = @"Apple wrote the addObserver / KVO notification API wrong in the first place and now requires developers to pass around pointers to fake objects to make up for the API deficicineces. You have to have one of these pointers per object, and they have to be internal and private. They serve no real value.";
-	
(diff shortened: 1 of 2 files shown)


adamgit:
There was a reason for calling initWithCoder - but it might not have been a good one :).

My guess: it was ... something ... to do with the fact initWithCoder is called by Xcode-Storyboards/NIBs, and that's the only case where it's literally impossible to provide the image in "init...". That's the main case where we have to catch it and warn the developer.

If you've a better way, go for it. The key thing is to make sure that Xcode-created instances correctly show up with the warning, as well as the lesser problem of coder-created-with-a-nil-SVGKImage-reference.

dgileadi:
Yeah, calling initWithCoder makes sense when considering that UIKit calls that method for image views in storyboard/NIB files. However it doesn't make as much sense to call it with a nil coder–calling the super method with a nil coder seems to create an invalid view, but avoiding calling the super method at all breaks putting image views in storyboard/NIB files. I suppose the current code is a decent compromise :)

Incidentally the whole reason why I modified SVGK*ImageView at all is because I'm using it in a storyboard and found that it didn't respect my positioning, sizing and other properties because the super initWithCoder wasn't originally called. In my case I'm using SVGKFastImageView which doesn't show the warning SVG, but unirregardless it was clear enough to me that I needed to provide an image programmatically. I think the warning message in the log helps.

PR #131, opened 2013-11-25

cbhl:
This fixes issue #106.


cbhl:
This is a resubmission of Pull Request #111, but targeting the 1.x branch instead of 1.1.0.

It has been modified to use a hard tab instead of spaces, to better match the surrounding code.

Issue #108, opened 2013-09-30
Image parse but not draw when using "use" 


flemingm:
Fixed bug in SVGKParserDefsAndUse.m  
Y was stored in X instead of Y.
and added log message to confirm parsing was correct.

But don't know how to fix drawing issue parsing Quebec Flag at:
http://en.wikipedia.org/wiki/File:Flag_of_Quebec.svg

---

The fix and the Log message in:
- (Node_) handleStartElement:(NSString *)name document:(SVGKSource_) SVGKSource namePrefix:(NSString_)prefix namespaceURI:(NSString_) XMLNSURI attributes:(NSMutableDictionary _)attributes parseResult:(SVGKParseResult *)parseResult parentNode:(Node_) parentNode
  {
  ...

if( [name isEqualToString:@"use"])
        {
            /*\* NB: must supply a NON-qualified name if we have no specific prefix here ! */
            SVGUseElement *useElement = [[[SVGUseElement alloc] initWithQualifiedName:qualifiedName inNameSpaceURI:XMLNSURI attributes:attributes] autorelease];

```
        [useElement postProcessAttributesAddingErrorsTo:parseResult]; // handles "transform" and "style"

        if( [attributes valueForKey:@"x"] != nil )
            useElement.x = [SVGLength svgLengthFromNSString:[((Attr*)[attributes valueForKey:@"x"]) value]];
        if( [attributes valueForKey:@"y"] != nil )
            useElement.y = [SVGLength svgLengthFromNSString:[((Attr*)[attributes valueForKey:@"y"]) value]];
        if( [attributes valueForKey:@"width"] != nil )
            useElement.width = [SVGLength svgLengthFromNSString:[((Attr*)[attributes valueForKey:@"width"]) value]];
        if( [attributes valueForKey:@"height"] != nil )
            useElement.height = [SVGLength svgLengthFromNSString:[((Attr*)[attributes valueForKey:@"height"]) value]];

        NSString* hrefAttribute = [useElement getAttributeNS:@"http://www.w3.org/1999/xlink" localName:@"href"];

        NSAssert( [hrefAttribute length] > 0, @"Found an SVG <use> tag that has no 'xlink:href' attribute. File is invalid / don't know how to parse this" );
        if( [hrefAttribute length] > 0 )
        {
            NSString* linkHref = [((Attr*)[attributes valueForKey:@"xlink:href"]) value];

[...]

PR #134, opened 2013-12-12
Update SVGKParserDefsAndUse.m

flemingm:
Fixed  useElement.y useElement.width, useElement.height assignments.


flemingm:
part of the fix for issue #108.

PR #136, opened 2013-12-18
Made project build for all architectures (include arm64) and memory issues

cjhanson:
Made project build for all architectures (include arm64)
Fixed some memory leaks that the clang analyzer pointed out

### DIFF
--- a/SVGKit-iOS.xcodeproj/project.pbxproj
+++ b/SVGKit-iOS.xcodeproj/project.pbxproj
@@ -1190,7 +1190,7 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				ALWAYS_SEARCH_USER_PATHS = NO;
-				ARCHS = "$(ARCHS_STANDARD_32_BIT)";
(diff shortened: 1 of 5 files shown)


adamgit:
Quick check: We had some problems in Xcode 5.0.0 where it didn't correctly link libraries which had 64bit code - have you re-built the library from clean, and checked you can import it into an app (e.g. the demo app) and the app runs correctly?

(from (vague!) memory: Xcode 5.0.0 was dropping a slice when it output the .a file, and so you ended up with a library that was missing 386 or similar - i.e. wouldn't run on simulator, or wouldn't run on non-64bit devices ... something silly like that)

adamgit:
NB: I found a few that were missing after this, and just did another commit that added them. I think we've got the lot now (I did a Clean to check)

cjhanson:
Great thanks :) This project is super helpful.

adamgit:
Then comment here: https://github.com/SVGKit/SVGKit/issues/140

(I want to find out what people are using it for, and compile some screenshots + testimonials in the new year)

cjhanson:
I will certainly do that as soon as my app is on the store :)

PR #144, opened 2014-02-03
Not include the unnecessary file (SVGKParser.h) in SVGKImage.h so app can contain only it.

rhcad:
Enable both non-ARC and ARC mode.
Fix the issue “+[NSCharacterSet SVGWhitespaceCharacterSet]: unrecognized selector sent to app's class.


rhcad:
I'd like to use only SVGKImage.h in my project to load and draw SVG files, so I suggested that no include SVGKParser.h in SVGKImage.h to avoid more file are included too.

Now, I have set the header search path as `../../thirdparty/SVGKit/Source/**`, so the first commit can be discard. But a issue as below appear about the second commit.

> What's the issue here? Why does it matter that SVGKParser.h is imported?

rhcad:
I confess the second commit is incorrect, but it works in my project, or a issue `TestView-SVG[40407:70b] +[NSCharacterSet SVGWhitespaceCharacterSet]: unrecognized selector sent to class 0x17f0ff0` will occur.

My [TouchVG](https://github.com/rhcad/TouchVG/tree/master/ios) library project (libTouchVG-SVG.a) uses SVGKImage.h to draw SVG files, but not uses libTouchVG-SVG.a. It can be built. 

[TestView](https://github.com/rhcad/TouchVG/tree/master/ios/tests/TestView) application project (TestView-SVG.app) uses libSVGKit-iOS.1.2.0pre.a and libTouchVG-SVG.a, but not uses SVGKImage.h directly. Then, the build issue occur.

Would you please help me to point out the reason? Thank you!

> This appears to be incorrect. Please check the Apple docs for "whitespaceAndNewlineCharacterSet". It does not satisfy the SVG specification (as quoted in the source comments).

The build and reappear way of [TouchVG](https://github.com/rhcad/TouchVG) (if you need it):
1. Check out TouchVG project containing [TouchVGCore](https://github.com/rhcad/TouchVGCore) and SVGKit submodules.
2. 'cd' to the 'ios' directory containing the file of 'build.sh', then type `./build.sh` to compile libraries.
3. Run TestView-SVG target.
4. Tap `GiPaintView SVG images` row, then the issue occurs.
`2014-02-05 00:50:33.655 TestView-SVG[40639:70b] +[NSCharacterSet SVGWhitespaceCharacterSet]: unrecognized selector sent to class 0x17f0ff0
2014-02-05 00:50:33.655 TestView-SVG[40639:70b] Fail to parse fonts.svg due to NSInvalidArgumentException`

rhcad:
This commit can be discarded too, I can force to change my projects as non-ARC mode when it uses SVGKit. I have tested the method (no better one probably) both non-ARC and ARC modes.

> This is a nice idea (both non-ARC and ARC), but it would force all developers to use a non-standard programming approach for all future code.

PR #149, opened 2014-02-16
nameUsedToInstantiate is undefined when not using ...GLOBAL_IMAGE_CACHE...

stephenhouser:
"#ifdef ENABLE_GLOBAL_IMAGE_CACHE_FOR_SVGKIMAGE_IMAGE_NAMED"
block. When ENABLE_GLOBAL_IMAGE_CACHE_FOR_SVGKIMAGE_IMAGE_NAMED
is not defined this property is not defined in the header file.

### DIFF
--- a/Source/SVGKImage.h
+++ b/Source/SVGKImage.h
@@ -44,7 +44,7 @@
 @class SVGKSource;
 @class SVGKParseResult;
 
-#define ENABLE_GLOBAL_IMAGE_CACHE_FOR_SVGKIMAGE_IMAGE_NAMED 1 // if ENABLED, then ALL instances created with imageNamed: are shared, and are NEVER RELEASED
+//#define ENABLE_GLOBAL_IMAGE_CACHE_FOR_SVGKIMAGE_IMAGE_NAMED 1 // if ENABLED, then ALL instances created with imageNamed: are shared, and are NEVER RELEASED
 
 @class SVGDefsElement;
 
--- a/Source/SVGKImage.m
+++ b/Source/SVGKImage.m
@@ -57,9 +57,9 @@
 @synthesize scale = _scale;
 @synthesize source;
 @synthesize parseErrorsAndWarnings;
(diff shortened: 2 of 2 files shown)


adamgit:
Good catch - the change to .m file is great, but I can't merge the change to .h file (you've reversed the behaviour of the library).

Please re-submit with ONLY the change to the .m file

(PS: if you don't want the shared image cache, it's much better to change the 1 to a 0, than to comment it out. If you comment it out, Xcode has no idea where it comes from, and cmd-clicking stops working for that #define)

stephenhouser:
Apologies for the changed #define line. I will resend with just the one change.

As to your P.S., the simple program below will illustrate the need for commenting out the #define. Both #if defined() and #ifdef will be TRUE if the object is defined as _anything_ (meaning 1 or 0) is irrelevant in the manner it is checked in the other code.

To fix that behavior will result in many more changes.. everywhere the #if is. If I have time, I'll take a stab as I need it for my project and would like to stay in sync. 

Then again, maybe a simple change to #undef would do it in one line and still have the Xcode shortcut...

/\* Does not matter if XXX is 0, 1, or "anything" both lines in main will print if it is defined. _/
#define XXX 0
#include &lt;stdio.h&gt;
int main(int argc, char *_argv) {
#ifdef XXX
        printf("#ifdef XXX\n");
#endif
#if defined(XXX)
        printf("#if defined(XXX)\n");
#endif
}

stephenhouser:
Nope, #undef won't do the trick. Ignore that hope.

adamgit:
The correct usage is:

#if THING_THAT_WAS_DEF'D

should work perfectly. If we have somewhere that's using #ifdef or #if
defined() ... those are definitely bugs!

stephenhouser:
All instances I found were using #ifdef. The later pull request changes them all.

adamgit:
Yep, that's great. GH crashed trying to merge, but I;ve just done it again,
seems to have taken this time.

PR #153, opened 2014-02-25
Fix unrecognized selector issue in [NS.. SVGWhitespaceCharacterSet].

rhcad:
The issue can be found in [TouchVG](https://github.com/rhcad/touchvg)'s TestView-SVG project:
TestView-SVG[3660:60b] +[NSCharacterSet SVGWhitespaceCharacterSet]: unrecognized selector sent to class 0x39e43da0

The build and reappear way of [TouchVG](https://github.com/rhcad/TouchVG) (if you need it):
1. Check out TouchVG project containing [TouchVGCore](https://github.com/rhcad/TouchVGCore) and SVGKit submodules.
2. 'cd' to the 'ios' directory containing the file of 'build.sh', then type `./build.sh` to compile libraries.
3. Run TestView-SVG target.
4. Tap `GiPaintView SVG images` row, then the issue occurs.
`2014-02-05 00:50:33.655 TestView-SVG[40639:70b] +[NSCharacterSet SVGWhitespaceCharacterSet]: unrecognized selector sent to class 0x17f0ff0
2014-02-05 00:50:33.655 TestView-SVG[40639:70b] Fail to parse fonts.svg due to NSInvalidArgumentException`

### DIFF
--- a/Source/Parsers/SVGKPointsAndPathsParser.m
+++ b/Source/Parsers/SVGKPointsAndPathsParser.m
@@ -196,8 +196,21 @@ inline BOOL SVGCurveEqualToCurve(SVGCurve curve1, SVGCurve curve2)
  */
 + (void) readWhitespace:(NSScanner*)scanner
 {
-    [scanner scanCharactersFromSet:[NSCharacterSet SVGWhitespaceCharacterSet]
+    [scanner scanCharactersFromSet:[SVGKPointsAndPathsParser SVGWhitespaceCharacterSet]
                         intoString:NULL];
+}
+
++ (NSCharacterSet *)SVGWhitespaceCharacterSet;
+{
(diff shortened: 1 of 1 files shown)


rhcad:
I used SVGKImage.h in ImageCache.mm, rather than *.m, so the extension function implemented in NSCharacterSet+SVGKExtensions.m can't be found?

adamgit:
Objective-C++ has some severe bugs, Apple doesnt really support it. I would
recommend using Objective-C

Apple has some hacks for ObjC to workaround linking bugs. Eg look for the
linker flag "-ObjC" - there might be an equivalent C++ flag.

And of course you need the ObjC flag in your own projects, as per the
install instructions.

rhcad:
Thank you for pointing out the reason, It works!

PR #154, opened 2014-02-27
Updated podspec so it will actually build when added as a pod to a project

amayers:
With these changes users can use `pod 'SVGKit', :git => 'https://github.com/SVGKit/SVGKit.git'` to install SVGKit as a Cocoapod. Ideally we would replace the outdated v1.0 that is in the Cocoapods spec repo, but to do that you would have to transition to tagging releases, the podspec works as is but it does throw a warning that it is just pointing to the 1.x branch instead of a tag. The Cocoapod's team doesn't accept pod specs with warnings.

Going forward on development with SVGKit here are a couple things to be aware of. 
1) If you add any new source files outside the current directory structure or add new sub directories you will need to update the pod spec's `s.ios.source_files` to include that path. 
2) CocoaLumberjack will be included as a dependency. This means it will use the latest published CocoaLumberjack version, and not the version you have been including in the project so far. This fixes cases where people add SVGKit to their app and they have a conflict since their app or another lib it uses had also included CocoaLumberjack.

### DIFF
--- a/SVGKit.podspec
+++ b/SVGKit.podspec
@@ -18,8 +18,10 @@ Pod::Spec.new do |s|
                  'Eric Man'        => 'Eric@eric-mans-macbook-2.local' }
   s.source   = { :git => 'https://github.com/SVGKit/SVGKit.git', :branch => "1.x" }
 
-  s.ios.source_files = 'Source/*{.h,m}', 'Source/**/*.{h,m}'
+  s.ios.source_files = 'Source/*{.h,m}', 'Source/DOM classes/**/*.{h,m}', 'Source/Exporters/*.{h,m}', 'Source/Parsers/**/*.{h,m}', 'Source/QuartzCore additions/**/*.{h,m}', 'Source/Sources/**/*.{h,m}', 'Source/UIKit additions/**/*.{h,m}', 'Source/Unsorted/**/*.{h,m}'
   s.libraries = 'xml2'
   s.framework = 'QuartzCore', 'CoreText'
+  s.dependency 'CocoaLumberjack'
+  s.prefix_header_file = 'XCodeProjectData/SVGKit-iOS/SVGKit-iOS-Prefix.pch'
   s.xcconfig = { 'HEADER_SEARCH_PATHS' => '$(SDKROOT)/usr/include/libxml2' }
 end


adamgit:
Does this work OK with lumberjack? Lumberjack was designed so that this connection is impossible, and they only recently changed it to make it work as an externally-linked library.

Apparently, we need to make a few-lines change to SVGkit to support that - I guess it works OK if you don't so long as you have the new version of Lumberjack?

amayers:
I haven't tested it to make sure that lumberjack is logging properly, but everything will now build and you can now use SVGKit when installed via Cocoapods. As for using SVGKit the normal way you have been (without Cocoapods), this change shouldn't hurt that since I'm only changing the podspec file.

adamgit:
Can you run the demo app and check it outputs something? It should spew out
a bunch of logging if you open any of the SVG's, direct to console by
default. There are no NSLog's in there - so if you see anything, it means
it's working :)

amayers:
The demo app in the SVGKit repo is unaffected by this change since it doesn't use Cocoapods. When I use Cocoapods to add SVGKit to an app I'm working on, only SVGKit's static library is included not the demo app. When using SVGKit in a test project to verify that the podspec is working I don't see any logging. However the only thing I am doing in that app is `SVGKFastImageView *svgFastImageView = [[SVGKFastImageView alloc] initWithSVGKImage:[SVGKImage imageNamed:@"Signature_of_Vishwanathan_Anand"]];` Would that typically log anything?

adamgit:
Hmm. Try loading the World Map from the demo project (drag/drop into your
project) - it generates some Warnings (because the file is missing SVG
info, so SVGkit has to guess the missing data).

I think it's called "WordMapEquirect-6.svg" or something like that

amayers:
I have to call `[SVGKit enableLogging];` in my project but after that it logs the same as your demo app does.

LeChatNoir69:
@amayers : so now, we can use SVGKit with Cocoapod ? Do you confirm that ? Thxs

amayers:
@LeChatNoir69 Yes you can now use Cocoapods to install SVGKit. However since the podspec that was submitted to the spec repo was created by some unknown 3rd party it causes a few issues. Ideally @adamgit would need to submit the updated podspec and convince the maintainers of the spec repo to replace the current one and give him access since he is the project's owner. However until then you will need to use this long form format to use it.

Instead of
`pod "SVGKit"`
use
`pod "SVGKit", :git => 'https://github.com/SVGKit/SVGKit.git', :branch => '1.x'`

LeChatNoir69:
Ok thank you. I'll give it a try.
Cheers

LeChatNoir69:
unfortunately, it doesn't work. Got this message : 

ArgumentError - Illformed requirement `":git => 'https://github.com/SVGKit/SVGKit.git' :branch => '1.x'"`

What's wrong ?

adamgit:
I guess you should ask on cocoapods support?

amayers:
@LeChatNoir69 I don't know what is causing that issue for you. I just copy/pasted the `pod "SVGKit", :git => 'https://github.com/SVGKit/SVGKit.git', :branch => '1.x'` line from above into a test project and I can pull down the pod fine. Maybe you have a fairly old version of Cocoapods and it doesn't recognize the `:branch => '1.x'` portion. I think I remember seeing that was added in the last few versions. Open terminal and run `sudo gem update` to update Cocoapods then try again. Beyond that there is something wrong on your end and not with the pod.

LeChatNoir69:
You're right. After a gem update, it's better.
I've this error now :
Pre-downloading: `SVGKit` from `https://github.com/SVGKit/SVGKit.git`, branch `1.x`
Just downloaded and checked out branch: 1.x from upstream /Users/LeChat/Library/Caches/CocoaPods/GitHub/8308a01fc1a5db5285df878f1b47aceacd4689a9
[!] The platform of the target `Pods` (iOS 6.0) is not compatible with `SVGKit (1.x)` which has a minimum requirement of iOS 7.0.

I don't think SVGKit requires iOS7 ?!
@adam : do you confirm ?

@amayers : Do you think I can input iOS7, make the pod install and reinput iOS6.

Thxs :)

LeChatNoir69:
Ok. I've change iOS6 to iOS7 in the pod file :
platform :ios, '7.0'

And everything is ok.
I assume that I can change deplyment target in the pod target to iOS6 because my app will be iOS6 compatible.

Thank you very much ! I'll now test if everything is ok ;)

adamgit:
SVGKit definitely does _not_ require iOS7.

Most of it works on iOS4 (should be "all" but I haven't tested an iOS4
device in a while).

LeChatNoir69:
@amayers : everything is ok. Big thank you for your help.
@adam, you should accept amayers request in order to avoid this iOS7 compatibility false message.

Cheers

adamgit:
Yep, I beat you to it by 2 days ;)

LeChatNoir69:
^^ Thxs :)

PR #168, opened 2014-05-16
Add CLANG_CXX_LANGUAGE_STANDARD and requires_arc to the CocoaPods specs.

rhcad:
Other suggestions:

``` ruby
s.version     = '1.2.0-alpha'
```

### DIFF
--- a/SVGKit.podspec
+++ b/SVGKit.podspec
@@ -2,16 +2,14 @@ Pod::Spec.new do |s|
   s.name        = 'SVGKit'
   s.version     = '1.x'
   s.license     = 'MIT'
-  s.platform    = :ios, '4.0'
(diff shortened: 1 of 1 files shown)


adamgit:
As far as I'm aware, this still runs on iOS 4.

c.f. the other threads, the consensus was that we'd bump the version number before we increase the requries iOS version (and probably switch to ARC at the same time?)

rhcad:
iOS 4 and SVGKit version can remain unchanged, but I think `gnu++11` and non-ARC should be specified.

adamgit:
OK, sounds good. I can't edit the commit, sorry - you'll have to resubmit.

I think we should move to new version very soon! :)

rhcad:
Moved to #169 .

PR #174, opened 2014-06-04
Fixed "rotate" transform type.

andmar1x:
Tested on attached file, which is displayed correctly in Google Chrome. https://www.dropbox.com/s/tqsx5nv01krncd5/test.svg

### DIFF
--- a/Source/DOM classes/Unported or Partial DOM/SVGElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGElement.m
@@ -319,9 +319,10 @@
 				{
 					CGFloat degrees = [[parameterStrings objectAtIndex:0] floatValue];
 					CGFloat radians = degrees * M_PI / 180.0;
-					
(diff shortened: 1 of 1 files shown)


adamgit:
How does this work? It appears to be doing the transformation twice :O

Also, does this work OK with our test image in the SVGKitDemo app - off the top of my head, one of the first few images from the top does lots of rotations and scales (it has text in the image saying so)

andmar1x:
I just commented out a few old lines and replaced them with equivalent (CGAffineTransformMakeTranslation to CGAffineTransformTranslate, CGAffineTransformMakeRotation to CGAffineTransformRotate and other).

> Also, does this work OK with our test image in the SVGKitDemo app - off the top of my head, one of the first few images from the top does lots of rotations and scales (it has text in the image saying so)

If you are talking about the "rotated-and-skewed-text" example, everything works fine.

PR #179, opened 2014-06-30
fix memory leak of NSDates

dpletter:
Loading any svg file into a SVGKFastImageView shows leaks of 2 NSDate objects from SVGKFastImageView when leak profiling in Instruments.  This change resolves those leaks.

### DIFF
--- a/Source/UIKit additions/SVGKFastImageView.m
+++ b/Source/UIKit additions/SVGKFastImageView.m
@@ -185,6 +185,8 @@
 	[self removeObserver:self forKeyPath:@"showBorder" context:internalContextPointerBecauseApplesDemandsIt];
     
 	self.image = nil;
+    self.startRenderTime = nil;
+    self.endRenderTime = nil;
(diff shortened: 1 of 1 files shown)


adamgit:
Please can you re-apply this to the current branch?

Use the front page, or point your git client at branch "2.x"

dpletter:
Ok, but should the readme be updated?

"
v1.x = current "in development" branch with latest changes, fixes, features
NB: this is now automatically selected in GitHub as the "default" branch when you visit SVGKit's project page
USE THIS FOR EVERYTHING UNLESS YOU HAVE A VERY GOOD REASON NOT TO!
"

adamgit:
Yes, it's wrong!

adamgit:
Closing because this is now applied to 2.x, and the README was fixed in https://github.com/SVGKit/SVGKit/commit/cd47ae95e57c9490a9238173855ecbe83d4aaf44

PR #182, opened 2014-07-02
add populateFromImage in initWithFrame method

j5136p1:
when you init it with initWithFrame method you get a observer error for key path layer because you never add a observer. I copied the call from initWithCoder. I think that should be the same behavior.


adamgit:
Please apply to 2.x branch! I don't want to make changes to 1.x branch unless they're super-critical.

PR #186, opened 2014-07-17
fix some memory leak; add pathForShapeInRelativeCoords in SVGEllipseElement.m

ArcRain:
Found some issue and repair them.

### DIFF
--- a/Source/DOM classes/Unported or Partial DOM/SVGEllipseElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGEllipseElement.m
@@ -50,14 +50,4 @@
     CGPathRelease(path);
 }
 
--(CALayer *)newLayer
-{
-	CGMutablePathRef path = CGPathCreateMutable();
-	CGPathAddEllipseInRect(path, NULL, CGRectMake(_cx - _rx, _cy - _ry, _rx * 2, _ry * 2));
-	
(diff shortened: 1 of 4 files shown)


Source/DOM classes/Unported or Partial DOM/SVGEllipseElement.m
```diff
@@ -43,6 +43,11 @@ - (void)postProcessAttributesAddingErrorsTo:(SVGKParseResult *)parseResult {
 	if( [[self getAttribute:@"r"] length] > 0 ) { // circle
 		self.ry = self.rx = [[self getAttribute:@"r"] floatValue];
 	}
+    
+    CGMutablePathRef path = CGPathCreateMutable();
+	CGPathAddEllipseInRect(path, NULL, CGRectMake(self.cx - self.rx, self.cy - self.ry, self.rx * 2, self.ry * 2));
+	self.pathForShapeInRelativeCoords = path;
+    CGPathRelease(path);
```
adamgit, line 50: If you make this change, you must also delete "newLayer".

Currently, your change has no effect!

ArcRain, line 50: What I want is to get "pathForShapeInRelativeCoords" without call "newLayer". So, I think we can delete "newLayer" when using my change.

ArcRain:
I have deleted  "newLayer" and commit it.

PR #191, opened 2014-08-09
Added the ability to load SVG images from relative paths.

dgileadi:
&lt;image&gt; elements should be able to be loaded from paths that are relative to the original SVG source. This implies that that SVGImageElement needs access to the SVGKSource that the SVG came from. This pull request adds the support, making the source available in SVGSVGElement and using it in SVGImageElement to load the image from the relative path.

Note that this pull request adds the file NSData+NSInputStream which comes from https://github.com/tipbit/tidbits and is MIT licensed. If that's an issue then I can easily recreate the file myself.


Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
```diff
@@ -81,7 +82,14 @@ - (CALayer *) newLayer
...
 	newLayer.frame = frame;
 	
-	NSData *imageData = [NSData dataWithContentsOfURL:[NSURL URLWithString:_href]];
+	NSData *imageData;
+	if( [_href hasPrefix:@"data:"] || [_href hasPrefix:@"http:"] )
+		imageData = [NSData dataWithContentsOfURL:[NSURL URLWithString:_href]];
+	else
+	{
+		NSInputStream *stream = [self.rootOfCurrentDocumentFragment.source sourceFromRelativePath:_href].stream;
+		imageData = [NSData dataWithContentsOfStream:stream initialCapacity:NSUIntegerMax error:nil];
```
adamgit: Please do something with the NSError parameter.

It is sufficient to DDLogError() it - but we mustn't silently drop errors!

adamgit:
Looks good. Great to see the SVGKSource stuff helping here.

I had one in-line comment (re an NSError being dropped). A couple of general points too:
1. Can you write a trivial (or find a public domain) SVG file that uses these features, for us to include in the SVGKitDemo apps? I want to guarantee that no-one breaks your fixes in future by accident! (has happened a few times with the less-used features)
2. License: appears compatible/same as ours? Maybe requires that an extra copyright line is added (Something like "Some parts Copyright BLAHBLAH" ?) https://github.com/SVGKit/SVGKit/blob/2.x/LICENSE
3. Not a problem, but -- Is the layoutLayer implementation necessary? I thought this was inherited from super already (so we don't need to keep re-implementing it). Is there a problem with the super version? (or am I misremembering it?)
4. I thought we already had code in the project that loads an NSData from NSInputStream using Apple's APIs. Some people have reported problems with it - if Apple's code is known to be buggy, should we be using this new code to replace all our instances of that?

adamgit:
Also, re: 1 above - I'll use that to verify this patch when merging. I'll try the file locally, check it crashes/fails to render, then run your partched version, check it renders correctly.

dgileadi:
Regarding your inline comment I fixed the code to check for and log any error. Thanks for catching that.

Regarding your general comments:
1. I created a file using a public-domain bitmap image and added it to the demo app.
2. Yes, the license appears the same. However I'm no lawyer and don't know what special things we may need to do to include it, if any.
3. This snuck in due to my obtuseness—I foolishly put a commit for upcoming <clipPath> support into the same branch, forgetting that it would also end up in the pull request. I've backed those changes out.
4. I searched the code base and didn't find anything to load an NSData from NSInputStream. Maybe I missed it somehow.

PR #192, opened 2014-08-11
Support for clip path

dgileadi:
Adds support for &lt;clipPath&gt; and the clip-path attribute. &lt;clipPath&gt; elements aren't rendered as part of the normal display but if they're referenced in a clip-path attribute then they clip the element that references them.

This implementation uses the CALayer's mask property to do the masking. This is not entirely correct since the definition of &lt;clipPath&gt; requires no alpha blending (except for anti-aliasing) while CALayer's mask property does support anti-aliasing, but most uses of &lt;clipPath&gt; don't contain alpha in their path definitions so this generally won't be a problem in practice. As a side note CALayer's mask property could also easily (and more correctly) be used to implement &lt;mask&gt;. I don't plan to do this however.

The Coins.svg sample in the demos has been changed to use &lt;clipPath&gt; to clip the images over a gray background.


adamgit:
NB: By Apple's (grudging) admission, CALayer masks are fundamentally broken on iOS. (This is why we have two versions of "SVGKImageView" -- to workaround the Apple bug!).

When you run this in Demo, are you forcing the demo to use SVGKLayeredImageView? Because SVGKFastImageView should (!) fail to render the mask correctly, according to Apple docs.

Ideally, can you add a log message when a file contains one of these clips that says something like:

DDLogWarn( @"clipPath found in SVG. Requires SVGKLayeredImageView, or similar CALayer rendering, to display correctly (workaround for Apple bug in CALayer .mask rendering)" );

(if you display that once per clipPath found during the parse, it would be OK - a little spammy, but not bad. Or  use the "errors and warnings" feature in SVGKParser, which enables you to aggregate a bunch of such warnings without logging them, and then at the end of parsing, you can output the warning log mesasge once)

EDIT: to be clear -- I just want something that prevents people logging Issues on GitHub "my image doesn't reneder properly! BUG!". Putting a warning both a) helps them fix it themself and b) saves us wasted support time chasing down problems somewhere in a 10,000 line SVG file :)

dgileadi:
I added the warning.

Interestingly my SVG is apparently using the default SVGKFastImageView and renders fine (you can try it out by running the Coins sample in the demo app; I stepped through to verify that it's using SVGKFastImageView). However it's a pretty simple SVG and masking could very well break when combined with fancy transforms, etc.

adamgit:
From memory: masking shouldn't work at all. Try opening the rainbow wing 
with fast image view (will need to tweak the demo app - IIRC at thr kast 
minute, it does a hardcoded check on the filename its loaded, before 
deciding which  to use)

dgileadi:
Hmm, I can't explain it working, but the Coins sample certainly is in both the simulator and on my iPod Touch.

In any case, with the warning now in place is there anything else I need to do to get this merged?

Source/DOM classes/Unported or Partial DOM/SVGClipPathElement.h
```diff
@@ -0,0 +1,39 @@
...
+#import <UIKit/UIKit.h>
+
+#import "SVGElement.h"
+#import "SVGElement_ForParser.h"
+
+#import "ConverterSVGToCALayer.h"
+#import "SVGTransformable.h"
+
+
+typedef enum SVG_CLIPPATHUNITS
```
adamgit: We need an interface "SVGUnitTypes" (in SVG-DOM folder, since this is simple to do right), which appears to be a header only, containing these constants.

We've been creating Enums where the Spec "implies" them. We have a MOSTLY consistent naming scheme :( (need to fix that!), where we've tried to make enums with same names/case as the Spec uses.

Looks to me like we should have either:

SVGUnitType.h:
   enum SVGUnitType { SVG_UNIT_TYPE_UNKNOWN ... }

or:

SVGUnitType.h:
   enum SVG_UNIT_TYPE { SVG_UNIT_TYPE_UNKNOWN ... }

...both of those look sensible to me. I think I prefer the first one, since ALL CAPS ususally means "constant" in C/objC.

What do you think?

Source/Parsers/Parser Extensions/SVGKParserSVG.m
```diff
@@ -33,6 +34,7 @@ - (id)init {
                           [SVGDescriptionElement class], @"description",
                           [SVGEllipseElement class], @"ellipse",
                           [SVGGElement class], @"g",
+                          [SVGClipPathElement class], @"clipPath",
```
adamgit, line 37: Is no additional parsing needed? I was expecting this class to also get some lines of code to parse-out "maskUnits" etc - are we already doing this?

Source/SVGKImage.m
```diff
@@ -626,7 +627,45 @@ - (CALayer *)newLayerWithElement:(SVGElement <ConverterSVGToCALayer> *)element
 	{
 		SVGUseElement* useElement = (SVGUseElement*) element;
 		childNodes = useElement.instanceRoot.correspondingElement.childNodes;
-	}
+    }
+    
+    /**
+     Special handling for clip-path; need to create their children
```
adamgit, line 619: If I understand the spec correctly ... and I freely admit I don't fully understand the clipping/masking! ... then this only applies to SVG, SYMBOL and IMAGE tags?

In which case, this code shouldn't be in SVGKImage, but instead in a helper class (I suggest making a new one called something like "SVGKHelperViewportTags" -- we can move some other bits of code there too) and explicitly invoked by each of those three classes when they're doing their "newLayer" methods.

(incidentally ... I just noticed that SVGHelperUtilties has illegal name. Yuck. We need to rename that one (I'll make an issue now, and someone can do it after we've merged all tehse recent chagnes)! (anything starting SVG is reserved by the spec...))

Source/DOM classes/Unported or Partial DOM/SVGClipPathElement.h
```diff
@@ -0,0 +1,39 @@
+/**
+http://www.w3.org/TR/SVG/masking.html#InterfaceSVGClipPathElement
```
adamgit: If you feel this is faithful to the spec (IMHO: looks like "yes, once the SVGClipPath enum is refactored into an SVGUnitTypes header") ... move this to the "SVG-DOM" folder too.

(the idea is to put all new classes in there, after making sure they are faithful, or 99% faithful. That way, we have a "hit-list" of classes that are old/mangled/need refactoring/cleaning ... the ones living in Unported-or-Partial-DOM)

Source/SVGKImage.m
dgileadi, line 619: The way I understood http://www.w3.org/TR/SVG/masking.html#ClipPathProperty the clip-path attribute can be on any graphical or container element. That's why I put the code here, since it needs to apply to just about everything.

Source/DOM classes/Unported or Partial DOM/SVGClipPathElement.h
dgileadi: Okay, I'll move it. I do believe it's faithful to the spec.

Source/SVGKImage.m
adamgit, line 619: Ah, I was looking at the "clip" property. http://www.w3.org/TR/SVG/masking.html#OverflowAndClipProperties

I think you're right.

Source/Parsers/Parser Extensions/SVGKParserSVG.m
dgileadi, line 37: No, since I didn't implement the &lt;mask&gt; element I don't have parsing for maskUnits. SVGClipPathElement does parse the clipPathUnits in its postProcessAttributesAddingErrorsTo: method, but I think that's about all that's needed.

dgileadi:
I've added support for &lt;clipPath&gt; when using renderInContext. It clips to a path when there's a single path as the mask; otherwise it renders the mask to a bitmap and clips using that. With this I'm able to render to a PDF context with clipped images appearing correctly.

adamgit:
1. I'm not seeing any clipping on the Coins file.
2. The Image tag is now broken; c.f. the "layered" image (second to last in the list). I'm not sure if it's this pull or the one before (that added the snowman).

dgileadi:
1. I created a pull request (#202) to make the clipping more obvious. If you see circular coin images on a red background then it's working; if you see squares then it isn't and please let me know.
2. I submitted a pull request (#201) to fix this; I suspect it was broken by the snowman merge. Sorry about that.

adamgit:
OK. It completely misunderstood what the image was showing, even with the red BG :).

If you've got the source images, howabout we do:
- green background
- red background on the coin images themselves

So then if you see any red = fail, but if it's all green = success ?

dgileadi:
I do have the source images, but it would be a bit of work to add a red background to the raster images, particularly since one of them's embedded in the SVG document. I'm inclined to let people just read Coins.svg to see what's going on, or just build their own clipPaths using Inkscape or whatever and have them work without drama.

I don't mind changing the sample to make the alarming red be some other more pacific color, though :)

PR #193, opened 2014-08-11
Change the public headers path to allow zero-config #import <SVGKit/SVGKit.h>

dgileadi:
Currently to use SVGKit in my iOS project I need to set up custom User Header Search Paths in Xcode to point to the Source folder in the SVGKit project. This pull request changes the location of the public headers that SVGKit generates to allow importing projects to import SVGKit headers using `#import &lt;SVGKit/SVGKit.h&gt;` without the need to set up any User Header Search Paths. It also changes the demo project accordingly.

### DIFF
--- a/Demo-iOS.xcodeproj/project.pbxproj
+++ b/Demo-iOS.xcodeproj/project.pbxproj
@@ -538,10 +538,6 @@
 				GCC_WARN_ABOUT_RETURN_TYPE = YES;
 				GCC_WARN_UNINITIALIZED_AUTOS = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				HEADER_SEARCH_PATHS = (
-					"../../iOS/**",
-					"../../Core/**",
-				);
 				IPHONEOS_DEPLOYMENT_TARGET = 4.3;
 				PROVISIONING_PROFILE = "";
 				"PROVISIONING_PROFILE[sdk=iphoneos*]" = "";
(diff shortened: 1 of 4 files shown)


adamgit:
Due to bugs in Xcode, this doesn't/didn't work.

Last time I checked, all versions up to and including xcode 5.0.x were 
broken. 5.1 and above may or may not be fixed.

It will APPEAR to work, but cause failures when your project rebuilds in 
other ways - eg when making an appstore build, or a release build.

If you can get this working on xcode 5.0+ and verified OK for all builds 
(sim, device, app store, enterprise, and release), great.

Otherwise I can't merge this. We've had too many problems in the past with 
it killing existing projects :(

MaddTheSane:
If SVGKit was made as an iOS framework, the include could work. However, making iOS frameworks is only supported on Xcode 6 and later, and will only work on iOS 8 and later.

dgileadi:
K, I'll take your word for it; I don't have the resources to test on all kinds of builds and I don't want to cause breakage for other folks.

adamgit:
Cool, that's good to know.

I keep meaning to try the fake frameworks ( as used by Facebook etc) but 
it's a lot of work to setup and test.

adamgit:
Here's a link to the "fake" frameworks (they're not fake, they're genuine,
but done using a workaround to the code missing from Xcode):

https://github.com/kstenerud/iOS-Universal-Framework

What I thought of doing, but never tried:
1. Create a REAL framework using the above
2. Package it as a SEPARATE PROJECT in the SVGKit repository (note that to
   re-build the REAL framework, you need to manually install an xcspec to your
   Xcode)
3. Whenever we do point releases, manually run 2 above and put a
   direct-link to the PRE-BUILT REAL framework on the front page of GitHub.

That way, devs are presented with:

a. The full source. Build it, use it live, but the headers require some
editing of header-paths
b. The pre-built "recent-ish" framework: you have the source (for debugging
- example code) but also the framework (for quick drag-n-drop install and
  run)
  c. Devs who want both A + B above have also got it, via item 2 above: they
  will need to install the xcspec, and suddenly they get best of all worlds.

PR #195, opened 2014-08-12
Support preserveAspectRatio for image elements

dgileadi:
Currently in SVGKit preserveAspectRatio is supported for &lt;svg&gt; elements but not for &lt;image&gt; elements. This pull request adds support for supporting preserveAspectRatio for raster images in &lt;image&gt; elements. It even supports the "slice" option which the &lt;svg&gt; support for preserveAspectRatio doesn't support.


Source/DOM classes/SVG-DOM/SVGFitToViewBox.h
```diff
@@ -19,4 +19,7 @@
 
 @property(nonatomic,retain) SVGAnimatedPreserveAspectRatio* preserveAspectRatio;
 
+-(double)aspectRatioFromWidthPerHeight;
+-(double)aspectRatioFromViewBox;
```
adamgit: This interface is defined by the spec. We cannot add to it.

If you need to add new methods, you'll need to create a custom class and/or extension header (that does NOT begin with the letters "SVG"; it can begin with "SVGK" instead)

dgileadi: It's not really needed; I'll remove.

Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
```diff
@@ -67,7 +67,9 @@ - (void)postProcessAttributesAddingErrorsTo:(SVGKParseResult *)parseResult {
 	_height = [[self getAttribute:@"height"] floatValue];
 
 	if( [[self getAttribute:@"href"] length] > 0 )
-	self.href = [self getAttribute:@"href"];
+        self.href = [self getAttribute:@"href"];
+    
+    [SVGHelperUtilities parsePreserveAspectRatioFor:self];
```
adamgit, line 76: Yuck. So the spec has "readonly attribute SVGAnimatedPreserveAspectRatio preserveAspectRatio;" shared between two interfaces that have no base interface. Bad spec :(.

LOL, sorry, didn't see the rest. Page was loading slowly. Nevermind!

dgileadi, line 76: I moved the parsing code to a new method in SVGHelperUtilities named parsePreserveAspectRatioFor:. I agree, copy/pasting the parsing code is unacceptable.  I'm not sure whether that location for the shared parsing code is acceptable; that's your call.

adamgit:
So ... this also refactors the "svg" tag to use the new code, right?

NB: Any class which is in the folder "SVG-DOM" I am keeping strictly to DOM spec. The classes outside that folder we inherited from earlier versions and do lots of bad things (that we try to clean up as we go along).

I probably made a couple of violations myself when setting up the SVG-DOM files - I couldn't see a clean workaround at the time, and there was a lot of code to write! - but I'm not letting any new ones in :). Hiding them via header/class extensions is OK.

NB: unfortunately, I can't merge this any more :(. I merged in changes from someone else that fixed embedded-SVG-images rendering, and that's conflicting with this now. GitHub won't tell me where/why :(.

Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
adamgit, line 76: Yes, sorry - my bad, when I wrote that I hadn't noticed you'd moved it.

We have a big TODO item of "remove all parsing code from the SVG\* classes" (because the spec doesn't allow it to be there). It's all got to move ... somewhere.

FYI: We've been having some good successes with making new "SVGKParser.....h" files for each tag / set of tags. If you look in the "Parsers/Parser Extensions" folder there's a few that people have added. Each one declares an array of tags it "takes owneship of parsing" for, and SVGKParser automatically picks them as needed.

...but that won't work here, because SVGSVG and SVGImage are completely different. So, yes, I think SVGHelperUtitliies is an OK place for now.

Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
```diff
@@ -93,9 +90,44 @@ - (CALayer *) newLayer
 		if( error )
 			DDLogError(@"[%@] ERROR: unable to read stream from %@ into NSData: %@", [self class], _href, error);
 	}
-	SVGImageRef image = [SVGImage imageWithData:imageData];
+    SVGImageRef image = [SVGImage imageWithData:imageData];
```
adamgit: I think this is a simple merge with the latest commit...

NB: latest commit renames SVGImage/SVGImageRef to more accurate names. SVGImage isn't allowed by the spec, I believe.

But in the end, latest commit is simply saying "if that SVGImageRef fails to load ... do something different". So I think your new code should insert easily/obviously, it's just all contained in braces for "if the image loaded OK" part.

?

dgileadi:
Merged 2.x into this branch so it should be good to go.

This does not refactor the SVG tag to use the new code. I originally went down the path of trying to make SVGSVGElement and SVGImageElement share the same code for preserving the aspect ratio during rendering, and came to the dismal conclusion that it just won't work due to how Apple renders images in CALayers—they always adjust the image to fit the bounding frame. The result of using the shared code was that images would appear twice as small as they should and offset to weird locations. So I share the code for parsing preserveAspectRatio but the layout code needs to be separate.

adamgit:
Yep. Manipulating the nested transforms to get everything in correct co-ordinate system at each child/grandchild/etc ... has been tortuous :). Especially when you get into the "and you can make an SVG tag inside an SVG tag! YAY!" world...

adamgit:
This commit has broken something in the image tag.

Before this commit, the image "imagetag-layered" partially (*) worked: you had a sun, a bird, and a tree (embedded from different sources).

After this commit, the sun has disappeared.

(*) = "partially" because there were a bunch of gradient objects that weren't rendering, apparently a separate bug in our gradient code at the moment.

dgileadi:
That almost sounds metaphorical :) In any case I'll investigate tomorrow.

On Aug 17, 2014, at 5:28 AM, adamgit notifications@github.com wrote:

This commit has broken something in the image tag.

Before this commit, the image "imagetag-layered" partially worked: you had
a sun, a bird, and a tree (embedded from different sources).

After this commit, the sun has disappeared.

dgileadi:
I created a pull request (#201) to fix this; it turns out that I need to apply my preserveAspectRatio code to the SVG-created image as well.

Assuming that the sun only rendered as a smiley face then the sample is fixed; if not then perhaps there's more going on.

adamgit:
Yep, looks good. The gradients in the image aren't renderinbg - but they never were, so that's a different problem ;)

PR #205, opened 2014-08-26
Add support for the textArea element from TinySVG 1.2.

dgileadi:
&lt;textArea&gt; adds a width, height and text wrapping over a standard &lt;text&gt;. It is not part of the standard SVG spec; it belongs to TinySVG 1.2. As such it may or may not be appropriate for SVGKit. However it's simple to implement on top of SVGTextElement and is useful for my needs so I'm creating this pull request in case it's wanted.


adamgit:
TinySVG doesn't use DOM, and seems to have left many class names undefined. But it looks like many/all of them are going to be added to SVG 2?

In which case, I'd prefer to name SVGTextAreaElement something like:

TinySVGTextAreaElement

...making it easy for us to have a separate SVGTextAreaElement if/when SVG 2 support is added.

(in anything except Obj-c, this wouldnt' be a problem. But without namespaces, we have to be paranoid :( ).

dgileadi:
Good call, I like the TinySVG prefix better. Changed.

PR #212, opened 2014-09-18
iOS 8 crash on SVG resize.

andy-lenox:
After switching to iOS 8, this would crash on any render and resize of all SVG's in my project. This fixes the issue for me.


adamgit:
Please include a log message eg ddloginfo in the catch. Can't have a silent catch! We need to authoritatively fix the underlying bug at some point.

adamgit:
No update in more than a year, closing this

PR #217, opened 2014-09-23
Add iPhone 6 and 6+ to PPI switch

dpletter:
Added iphone 6 and 6+ to the switch, updated default to be the iphone 6 PPI, assuming that future devices will be at least that high, and updated URL of reference page in comments, since the old page was deleted from wikipedia.

### DIFF
--- a/Source/DOM classes/SVG-DOM/SVGLength.m
+++ b/Source/DOM classes/SVG-DOM/SVGLength.m
@@ -126,8 +126,8 @@ static float cachedDevicePixelsPerInch;
 
 +(float) pixelsPerInchForCurrentDevice
 {
-	/** Using this as reference: http://en.wikipedia.org/wiki/List_of_displays_by_pixel_density#Apple
-	 */
+	/** Using this as reference: http://en.wikipedia.org/wiki/Retina_Display 
+      */
 	
 	size_t size;
 	sysctlbyname("hw.machine", NULL, &size, NULL, 0);
@@ -146,10 +146,13 @@ static float cachedDevicePixelsPerInch;
 	|| [platform hasPrefix:@"iPhone6"])
 		return 326.0f;
 	
+    if([platform hasPrefix:@"iPhone7"]) //iphone 6 and 6+
+        return 401.0f;
+    
 	if( [platform hasPrefix:@"iPhone"]) // catch-all for higher-end devices not yet existing
 	{
 		NSAssert(FALSE, @"Not supported yet: you are using an iPhone that didn't exist when this code was written, we have no idea what the pixel count per inch is!");
-		return 326.0f;
+		return 401.0f;
 	}
 	
 	if( [platform hasPrefix:@"iPod1"]


newmaniese:
The iPhone 6 still has a PPI of 326, it is the iPhone 6 Plus that is 401. We should check for each, instead of this.

nomothetis:
Is there a timeline for a 2.0.0 release?

adamgit:
I'm not sure you'd want a 2.0.0 release, it's not necessarily a good thing
:). All it means is "here's a point where the code won't change - no speed
improvements, no fixes, no more features".

If we hit 100% coverage of the spec, I'd push for a 2.0 release on
principle (or if we've already got a 2.x by then, I'd push for a 3.0!). But
at the moment no-one's contributing the missing pieces (mostly: animated
SVGs, plus a few others, including a lot of small niche stuff that doesn't
seem to appear in real-world SVG files - but we should still support
anyway) ... so that seems a long way off.

adamgit:
NB: the latest HEAD is always higher quality and more stable than anything
else you can get. Every commit and pull request gets tested before we
accept it.

There are occasional mistakes, things slip through - but we tend to fix
them within hours.

nomothetis:
Thanks for the quick response!

What we were looking for was a stable pod release that we could point to; we assumed 2.0.0 would be the next version. Ultimately, we were hoping to have the iPhone 6 fixes without having to fork the project and create our own internal pods. I take it that's not in cards?

Incidentally, does this mean SVGKit does not use semantic versioning? Because it seems like speed improvements and bugfixes could appear in 2.0.x releases, accessible via cocoapods with the '~> 2.0' directive.

adamgit:
The pod that points to 2.x branch is what you want, I suspect?

This is stable - but not frozen.

Because we haven't got 100% spec yet, and we have bugs and svgs that don't 
render, a frozen build is bad for most people.

(I did work on a project where we wanted a frozen build - we directly 
embedded SVGKit. This is the safest way to take a frozen version if that's 
what you need)

Re: semantic versioning - this is a small project with few contributors. We 
have no control over what gets added or fixed- all pull requests are 
gratefully received! - apart from rejecting submissions.

Managing semantic versioning in that situation seems.to be super hard. It 
takes more time than anyone has to spare - either you end up making many 
versions (which is a huge waste of time we don't have), or you make none. 
We are tending towards the latter :) simply because it leaves more time for 
actual.changes.

(I'd love to see semantic versioning at some point in future, when theres 
enough volunteers/ time to majntain it)

adamgit:
Also, I think it would make sense to do a 2.0 release if / when someone 
updates the whole codebase to remove all the old backwards compatible stuff.

Ie:
- make it ARC only
- remove all @synthesize calls
- use ios7/8 code and methods by preference, only keeping ios4/5/6 methods 
  in #if blocks

But no-one has volunteered for those tasks yet. It could be a year or more 
before someone chooses to.

nomothetis:
I'm confused…the CocoaPods/Specs repo seems to only have 0.0.1, 1.0 and 1.2.0. Did I miss the one that points to the 2.x branch?

And I totally understand about the maintaining semantic versioning and general support of the project. I'm working on some tools to help with versioning, but it might be a long time before they're available. Thanks for your help.

adamgit:
Anything in the cocoa pods repo is out of our control. There are plenty of 
problems with that, but you'll have to raise bug reports with them 
directly, sorry.

PR #222, opened 2014-09-26
add prefix SVGK to base objects

tibo:
I had a conflict between my projects classes and classes from the library (the Comment class to be precise) so I've made a fork of it to add the SVGK prefix to most of the DOM objects.

this involve some big changes to be able to include it to an existing project but in case you find this useful feel free to merge it to the main repository.


adamgit:
It's nice, but this prevents you from using any other DOM library (those
names are technically illegal, I believe - DOM requires you use precise
class names). I'd prefer to split the SVGKit project into two libraries - a
DOM and SVGKit itself - and then people can swap in/out any DOM
implementation they prefer.

MaddTheSane:
It seems that the Comment class is causing all sorts of misbehavior: there's a conflicting struct in OS X that makes it impossible for SVGKit to be used in OS X Swift code, and it can't be built with projects that use modules either due to me using some preprocessing trickery.

adamgit:
Yep - as per our previous brainstorming on this, I'm happy to merge one-off 
cases for situations like that where Apple has screwed up (because Devs can 
easily handle one off cases when switching to a new DOM).

But this fork makes it a major pain to switch DOM which I feel would be a 
major loss.

tibo:
No problem guys. 
I just wanted to share the problem I had and the solution I ended to. 
I understand this is not optimal in many cases

adamgit:
To be clear: I appreciate it, and its the best quick fix we have today (I 
think?).

tibo:
don't worry I totally get that!
this is indeed the easiest fix I've found. I don't have any other solution in mind, except maybe change the name of the classes in the project I'm working on...

adamgit:
I've posted an SO question, in the hope someone else has a neater / simpler way to achieve this (or a tool / script for doing it automagically?)

http://stackoverflow.com/questions/26929576/how-to-make-an-obj-c-library-globally-renameable-every-class-exported-symbol

adamgit:
For what it's worth, I tried using the aliasing trick mentioned in SO'com answers - but Apple's implementation is broken in Xcode 6, Xcode 7 - can't use it.

PR #228, opened 2014-10-20
Fix KVO-related crashes in SVGKFastImageView when setting the .image property

dgileadi:
I'm creating an SVGKFastImageView and setting its .image property instead of passing a SVGKImage to its initializer. When the SVGKImageView is later deallocated I get KVO-related crashes, saying it couldn't remove an observer that was never added. This patch fixes the issue for me on iOS 8, although based on the code I suspect the same problems would happen on earlier versions of iOS. The fix is to register all the observers in setImage instead of in populateFromImage, so that they're always registered when dealloc tries to unregister them. The dealloc method also changes to not call setImage (which would re-register them again), and instead releases the _image field directly.

The other change that is required for me to not crash on iOS 8 is to un-comment removing the "DOMTree.viewport" observer in SVGKImage, but that's not part of this pull request because I saw there's a separate issue for it.


Source/UIKit additions/SVGKFastImageView.m
```diff
@@ -184,7 +177,12 @@ - (void)dealloc
 	[self removeObserver:self forKeyPath:@"tileRatio" context:internalContextPointerBecauseApplesDemandsIt];
 	[self removeObserver:self forKeyPath:@"showBorder" context:internalContextPointerBecauseApplesDemandsIt];
     
-	self.image = nil;
+    if (_image) {
+        [_image removeObserver:self forKeyPath:@"size" context:internalContextPointerBecauseApplesDemandsIt];
+    }
+    
+    [_image release];
```
MaddTheSane, line 184: You might want to add `self.image = nil;` back in, although doing so would purely be a stylistic choice.

dgileadi, line 184: Not a stylistic choice at all. Setting self.image to nil would call [self setImage:nil] thanks to property rewriting, and in doing so would re-register all the observers.

dgileadi:
Just a note that I've updated this pull request with the latest merges, in case anyone's interested in pulling it.

PR #229, opened 2014-10-20
Fix iOS 8 bug with assert fail on infinity CGFloat

requilence:
iOS 8 produce 'inf' CGFloat instead of CGFLOAT_MAX on some CGPath (some bezier paths). But fails when try to assert this CGFloat as a valid float.
"Assertion failed: (CGFloatIsValid(x) && CGFloatIsValid(y)), function void CGPathMoveToPoint(CGMutablePathRef, const CGAffineTransform *, CGFloat, CGFloat), file Paths/CGPath.cc, line 254."

Final iOS 8.1 SDK and XCode 6.1 released but this bug remains. So there is a need for workaround.
This fix replace inf with CGFLOAT_MAX and -inf to -CGFLOAT_MAX in the returned coordinates by quartz

### DIFF
--- a/Source/QuartzCore additions/CGPathAdditions.m
+++ b/Source/QuartzCore additions/CGPathAdditions.m
@@ -57,7 +57,7 @@ void applier (void *info, const CGPathElement *element) {
 	CGFloat x = fixInfinity(pathInfo->offX);
 	CGFloat y = fixInfinity(pathInfo->offY);
     
-	const CGPoint *points = fixPointsInfinity(element->points);
+	const CGPoint *points = fixPointsInfinity(element);
 	
 	switch (element->type) {
 		case kCGPathElementMoveToPoint:
@@ -102,7 +102,7 @@ void applyPathTranslation (void *info, const CGPathElement *element) {
 	CGFloat x = fixInfinity(pathInfo->offX);
 	CGFloat y = fixInfinity(pathInfo->offY);
 	
-	const CGPoint *points = fixPointsInfinity(element->points);
+	const CGPoint *points = fixPointsInfinity(element);
 	
 	switch (element->type) {
 		case kCGPathElementMoveToPoint:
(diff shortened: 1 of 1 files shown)


Source/QuartzCore additions/CGPathAdditions.m
```diff
@@ -15,14 +15,31 @@
...
 
+CGFloat fixInfinity(CGFloat inputFloat){
+    if(inputFloat>CGFLOAT_MAX) inputFloat=CGFLOAT_MAX;
+    if(inputFloat<(-1)*CGFLOAT_MAX) inputFloat=(-1)*CGFLOAT_MAX;
+    return inputFloat;
+}
+
+CGPoint *fixPointInfinity(CGPoint *inputFloat){
+    int i;
+    for (i = 0; i < sizeof(inputFloat); ++i)
```
MaddTheSane: This doesn't look right…

Source/QuartzCore additions/CGPathAdditions.m
```diff
@@ -15,14 +15,31 @@
 	CGFloat offY;
 } PathInfo;
 
+CGFloat fixInfinity(CGFloat inputFloat){
+    if(inputFloat>CGFLOAT_MAX) inputFloat=CGFLOAT_MAX;
+    if(inputFloat<(-1)*CGFLOAT_MAX) inputFloat=(-1)*CGFLOAT_MAX;
+    return inputFloat;
+}
+
+CGPoint *fixPointInfinity(CGPoint *inputFloat){
```
MaddTheSane: If this is iterating over a C array, you probably want to pass the size of the array.

requilence:
@MaddTheSane Thank you. Forget about that thing. Fixed by switching of CGPathElementType.

adamgit:
Ouch, so that's what was going wrong. Good catch!

adamgit:
For bonus points, do you have a small SVG which shows this problem, and you can share / make public domain that we can add to the Demo project?

(Apple might change this again in future, it would be great to have a unit test to help in future)

PR #231, opened 2014-10-28
Change Node to make firstChild, lastChild, previousSibling and nextSibling be calculated properties

dgileadi:
The current implementation of Node never sets the firstChild, lastChild, previousSibling and nextSibling properties, meaning that they don't work. The DOM spec says that these properties are readonly. This makes them good candidates for calculated properties, based on the childNodes list. SVGKit already has a class, SVGElementInstance, that implements these as calculated properties. This pull request copies the code for these properties from SVGElementInstance.m into Node.m, and changes Node+Mutable.h to not make them readwrite.

The motivation for this change is the scenario where the SVG DOM is queried and modified, for example for apps that support minor edits of the SVG.

The other option for changing firstChild, lastChild, previousSibling and nextSibling would be to keep them as real properties and to make methods like insertBefore, replaceChild, removeChild and appendChild update them as appropriate. However I believe that this approach would be harder to maintain, and easier for them to get into an inconsistent state.

### DIFF
--- a/Source/DOM classes/Core DOM/Node+Mutable.h
+++ b/Source/DOM classes/Core DOM/Node+Mutable.h
@@ -10,10 +10,6 @@
 @property(nonatomic,readwrite) DOMNodeType nodeType;
 @property(nonatomic,assign,readwrite) Node* parentNode;
 @property(nonatomic,retain,readwrite) NodeList* childNodes;
-@property(nonatomic,assign,readwrite) Node* firstChild;
-@property(nonatomic,assign,readwrite) Node* lastChild;
-@property(nonatomic,assign,readwrite) Node* previousSibling;
-@property(nonatomic,assign,readwrite) Node* nextSibling;
 @property(nonatomic,retain,readwrite) NamedNodeMap* attributes;
 
 @property(nonatomic,assign,readwrite) Document* ownerDocument;
--- a/Source/DOM classes/Core DOM/Node.m
+++ b/Source/DOM classes/Core DOM/Node.m
@@ -187,6 +187,52 @@
 
 #pragma mark - Official DOM method implementations
 
+-(Node *)firstChild
+{
+    if( [self.childNodes length] < 1 )
+        return nil;
+    else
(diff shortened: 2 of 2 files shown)


Source/DOM classes/Core DOM/Node+Mutable.h
```diff
@@ -10,10 +10,6 @@
 @property(nonatomic,readwrite) DOMNodeType nodeType;
 @property(nonatomic,assign,readwrite) Node* parentNode;
 @property(nonatomic,retain,readwrite) NodeList* childNodes;
-@property(nonatomic,assign,readwrite) Node* firstChild;
-@property(nonatomic,assign,readwrite) Node* lastChild;
-@property(nonatomic,assign,readwrite) Node* previousSibling;
-@property(nonatomic,assign,readwrite) Node* nextSibling;
```
MaddTheSane, line 16: I'd suggest putting these back as `readonly` properties, like so:

```
@property(nonatomic,assign,readonly) Node* firstChild;
@property(nonatomic,assign,readonly) Node* lastChild;
@property(nonatomic,assign,readonly) Node* previousSibling;
@property(nonatomic,assign,readonly) Node* nextSibling;
```

dgileadi, line 16: Why? They’re already defined as readonly in Node.h; there’s no need to duplicate that definition in Node+Mutable.h.

MaddTheSane, line 16: Ah, my mistake. I thought this header was `Node.h`.

adamgit:
NB: there's quite a few bits of the DOM implementation that - like this - are overly simplistic (generally: not fully implemented).

I rushed through the absolute minimum we needed to get SVG parsing up and running, and I've tended to go back and add/fix implementations whenever I find myself needing a bit of DOM feature on a live project.

(and, as noted in other Issues, I'd love for our DOM implementation to be soft-linked, so that you could swap it out for alternative implementations at will. Fix some of those annoying name conflict problems!)

PR #232, opened 2014-11-04
Fix issue with editing/reloading XML

dgileadi:
When XML is saved and reloaded, images fail to display. This patch fixes the issue twice.

The issue was caused by the fact that SVGKImage.newLayerWithElement calls `[layer setNeedsDisplay]` if the layer has any child nodes. This apparently keeps layers with image content from displaying the image (thanks Apple). Normally <image> elements shouldn't have any whitespace, but Node.appendXMLToString adds newlines after every starting element, and always creates separate ending element tags. This is enough to create a text child within the <image> element.

There are two fixes:
1) Don't append newlines when outputting XML. Existing whitespace in Text children is already preserved and is sufficient.
2) Only count child nodes that create layers when deciding whether to return early in SVGKImage.newLayerWithElement; this prevents whitespace nodes from affecting the behavior.

### DIFF
--- a/Source/DOM classes/Core DOM/Node.m
+++ b/Source/DOM classes/Core DOM/Node.m
@@ -575,7 +575,7 @@
 		case DOMNodeType_DOCUMENT_NODE:
 		case DOMNodeType_ELEMENT_NODE:
 		{
-			[outputString appendString:@">\n"];
+			[outputString appendString:@">"];
 		}break;
 			
 		case DOMNodeType_DOCUMENT_TYPE_NODE:
@@ -638,7 +638,7 @@
 		case DOMNodeType_DOCUMENT_NODE:
 		case DOMNodeType_ELEMENT_NODE:
 		{
-			[outputString appendFormat:@"</%@>\n", self.nodeName];
+			[outputString appendFormat:@"</%@>", self.nodeName];
 		}break;
 			
 		case DOMNodeType_DOCUMENT_TYPE_NODE:
--- a/Source/SVGKImage.m
+++ b/Source/SVGKImage.m
@@ -658,7 +658,7 @@ static NSMutableDictionary* globalSVGKImageCache;
 	 
 	 (parent may have to change its size to fit children)
 	 */
-	NSUInteger childCount = 0;
+	NSUInteger sublayerCount = 0;
 	for (SVGElement *child in childNodes )
 	{
 		if ([child conformsToProtocol:@protocol(ConverterSVGToCALayer)]) {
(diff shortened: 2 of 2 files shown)


Source/DOM classes/Core DOM/Node.m
```diff
@@ -575,7 +575,7 @@ -(void) appendXMLToString:(NSMutableString*) outputString availableNamespaces:(N
 		case DOMNodeType_DOCUMENT_NODE:
 		case DOMNodeType_ELEMENT_NODE:
 		{
-			[outputString appendString:@">\n"];
```
adamgit, line 578: From memory, this was because user-generated SVG rarely has whitespace manually added (that's a lot of hassle to do by hand!), and so our output SVGs were horrible to work with.

I guess what we really need is an OPTIONAL "pretty printer" output for that case. Maybe add another library (must be a good XML formatter library that's open-source and ObjC - maybe even one in the Apple libraries?)

Source/SVGKImage.m
```diff
@@ -670,12 +667,17 @@ - (CALayer *)newLayerWithElement:(SVGElement <ConverterSVGToCALayer> *)element
...
 				continue;
-            }
+			}
 			
+			childCount++;
 			[layer addSublayer:sublayer];
 		}
 	}
 	
+	if ( childCount < 1 ) {
```
adamgit: This change is subtle and yet important.

Please add a code-comment  approximately here something like:

"If none of the child nodes return a CALayer, we're safe to early-out here. But there are some nodes (e.g. whitespace) that appear as children, but don't affect the CALayer stack, and we need to filter them before making that decision"

Also, would be much better to rename "childCount" to "childrenWhichGenerateLayersCount" or similiar: make it explciit that it's a special count, otherwise someone is bound to replace it with node.childCount or similar in future, and break this :)

adamgit: Or maybe even better "numSublayersAdded" -- that's what we really care about: "did we do any work here, while iterating, or was it a null op (in which case: we can early-out)"

Source/DOM classes/Core DOM/Node.m
dgileadi, line 578: libxml2 has an xmlwriter, and there are several good ObjC XML libraries out there. However I think that switching to one would be a rather major change and probably not appropriate for this pull request.

I understand that removing the newlines is a hassle for user-generated SVG, but on the other hand the current behavior can be considered broken. With the current behavior the generated SVG grows each time it's saved, in addition to the problem that the SVGKImage change fixes. I think the change is worth the hassle.

My opinion is that the change to remove newlines should stand as-is, and in the future a separate change could switch to a different XML library or to libxml2's xmlwriter.

adamgit, line 578: Agreed. Line comment was more a note for future / addition, rather than a change to this commit.

I've not looked at libxml2's writer in a long time, but I vaguely remember it's very basic / buggy. Might be good enough though!

PR #233, opened 2014-11-05
Supporting svg fill-rule. Renaming the Comment to SVGComment.

krzysiek84:


adamgit:
Looks great, but we cannot rename Comment - you are breaking both APIs if you do that!

(It's illegal in SVG to name anything "SVGfoo", and it's illegal in DOM to name "Comment" anything except "Comment")

Please remove the Comment rename. If you want to workaround the name-clash, please provide a separate commit. What we're looking for is something that renames _all_ DOM classes, and is optional (!), and gives the developer a way of controlling the rename process (because whateer you rename to, I can guarantee it will classh sooner or later :( ).

Source/DOM classes/Unported or Partial DOM/SVGElement.m
```diff
@@ -539,7 +539,10 @@ -(NSString*) cascadedValueForStylableProperty:(NSString*) stylableProperty
 			
 			if( parentElement == nil
 			   || [parentElement isKindOfClass:[SVGSVGElement class]] )
-				return nil; // give up!
+                
+                //FIXME: not sure if this is correct,
+                //if we did not found any style in any of parent nodes, take one from root
+                return [self.rootOfCurrentDocumentFragment.style getPropertyValue:stylableProperty];
```
adamgit, line 545: How is this different? If you haven't found anything by going through the full tree, then ... how would looking at the root make a difference?

Off the top of my head, I don't know why we refuse to accept styles defined on "SVGSVGElement" (which will be the exact same as self.rootOfCurrentDocumentFragment) - is this in the spec, maybe?

krzysiek84, line 545: SVG I have has following format:

`<svg ... style="fill-rule:evenodd; ">
    <g>
      <g>
       <path ...>
      </g>
   </g>
</svg>`

When testing the lib with that svg it was not detecting correctly the style. But it renders fine in Firefox so i think the format is ok.

I have put the code with //Fixme, as maybe there is better fix for it. But did not put too much time into better solution.

krzysiek84:
Going to create new pull request without the rename and //Fixme.

PR #237, opened 2014-11-13
Infer width and height from viewBox

chetbox:
Some SVG editors and exporters do not include the `<svg>` "width" and "height" attributes.
e.g.

``` xml
<svg version="1.1"
         xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
         x="0px"
         y="0px"
     viewBox="0 0 104 144">
...
```

SVGKit currently will not render these files. We found that Adobe Illustrator exports files like this if you check the "Responsive" option when exporting.

Additionally, some editors and exporters set percentages as their "width" and "height" values to indicate a relative size at which to display the image to a browser. e.g.

``` xml
<svg version="1.1"
         xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"
         x="0px"
         y="0px"
         width="100%"
         height="100%"
     viewBox="0 0 104 144">
...
```

SVGKit currently crashes when rendering these files. We've found that Inskscape saves files like this when saving as "SVG Plain".

This change ensures that:
- Percentage `width`s and `height`s are ignored
- If no `width` and `height` attributes are present, we simply grab the values from the `viewBox` attribute.

### DIFF
--- a/Source/DOM classes/SVG-DOM/SVGLength.m
+++ b/Source/DOM classes/SVG-DOM/SVGLength.m
@@ -112,6 +112,18 @@ static float cachedDevicePixelsPerInch;
 	return result;
 }
 
++(SVGLength*) svgLengthFromNumber:(float) f
+{
+    CSSPrimitiveValue* pv = [[[CSSPrimitiveValue alloc] init] autorelease];
+
+    pv.pixelsPerInch = cachedDevicePixelsPerInch;
+    [pv setFloatValue:CSS_NUMBER floatValue:f];
+
+    SVGLength* result = [[[SVGLength alloc] initWithCSSPrimitiveValue:pv] autorelease];
+
+    return result;
+}
+
 -(float) pixelsValue
 {
 	return [self.internalCSSPrimitiveValue getFloatValue:CSS_PX];
--- a/Source/DOM classes/Unported or Partial DOM/SVGSVGElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGSVGElement.m
@@ -139,37 +139,50 @@
 		self.height = nil; // i.e. undefined
 	else
 		self.height = [SVGLength svgLengthFromNSString:[self getAttribute:@"height"]];
-	
-	/* set the frameRequestedViewport appropriately (NB: spec doesn't allow for this but it REQUIRES it to be done and saved!) */
-	if( self.width != nil && self.height != nil )
-		self.requestedViewport = SVGRectMake( 0, 0, [self.width pixelsValue], [self.height pixelsValue] );
-	else
-		self.requestedViewport = SVGRectUninitialized();
-	
-	
-	/**
-	 NB: this is VERY CONFUSING due to badly written SVG Spec, but: the viewport MUST NOT be set by the parser,
-	 it MUST ONLY be set by the "renderer" -- and the renderer MAY have decided to use a different viewport from
-	 the one that the SVG file *implies* (e.g. if the user scales the SVG, the viewport WILL BE DIFFERENT,
-	 by definition!
-	 
-	 ...However: the renderer will ALWAYS start with the default viewport values (that are calcualted by the parsing process)
-	 and it makes it much cleaner and safer to implement if we have the PARSER set the viewport initially
-	 
-	 (and the renderer will IMMEDIATELY overwrite them once the parsing is finished IFF IT NEEDS TO)
-	 */
-	self.viewport = self.requestedViewport; // renderer can/will change the .viewport, but .requestedViewport can only be set by the PARSER
-	
+
+
 	if( [[self getAttribute:@"viewBox"] length] > 0 )
 	{
 		NSArray* boxElements = [[self getAttribute:@"viewBox"] componentsSeparatedByString:@" "];
 		
 		_viewBox = SVGRectMake([[boxElements objectAtIndex:0] floatValue], [[boxElements objectAtIndex:1] floatValue], [[boxElements objectAtIndex:2] floatValue], [[boxElements objectAtIndex:3] floatValue]);
-	}
+
+        /**
+         Infer width and height from viewBox if not specified
+         */
+
+        if (self.width == nil)
+            self.width = [SVGLength svgLengthFromNumber:_viewBox.width];
+
+        if (self.height == nil)
+            self.height = [SVGLength svgLengthFromNumber:_viewBox.height];
+    }
 	else
 	{
 		self.viewBox = SVGRectUninitialized(); // VERY IMPORTANT: we MUST make it clear this was never initialized, instead of saying its 0,0,0,0 !		
 	}
+
+
+    /* set the frameRequestedViewport appropriately (NB: spec doesn't allow for this but it REQUIRES it to be done and saved!) */
+    if( self.width != nil && self.height != nil )
+        self.requestedViewport = SVGRectMake( 0, 0, [self.width pixelsValue], [self.height pixelsValue] );
+    else
+        self.requestedViewport = SVGRectUninitialized();
+
+
+    /**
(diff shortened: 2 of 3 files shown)


adamgit:
Off the top of my head, you cannot take width and height from viewbox - 
IIRC its illegal.

I'm checking now in spec, but this stuff is subtle.

adamgit:
A good svg to test with is the world map - it has a negative view box origin (which is horrible, but legal!).

Also, do you have some test SVGs you can release as public domain and include in thedemo project to demonstrate your changes working? Make sure no one breaks this in future...

Source/DOM classes/SVG-DOM/SVGLength.m
```diff
@@ -112,6 +112,18 @@ +(SVGLength*) svgLengthFromNSString:(NSString*) s
 	return result;
 }
 
++(SVGLength*) svgLengthFromNumber:(float) f
```
adamgit, line 115: Shouldnt this be part of the existing method?

Source/DOM classes/Unported or Partial DOM/SVGSVGElement.m
```diff
@@ -171,26 +162,44 @@ We would need to put extra (NON STANDARD) properties on SVGDocument, for the "vi
...
         }
 		_viewBox = SVGRectMake([[boxElements objectAtIndex:0] floatValue], [[boxElements objectAtIndex:1] floatValue], [[boxElements objectAtIndex:2] floatValue], [[boxElements objectAtIndex:3] floatValue]);
-	}
+
+        /**
+         Infer width and height from viewBox if not specified
+         */
+
+        if (self.width == nil)
+            self.width = [SVGLength svgLengthFromNumber:_viewBox.width];
```
adamgit, line 171: Better than this would be to add a non-spec method e.g. "-(BOOL) canConvertToAbsoluteValue" which returns false when the length is a percentage AND there is no parent-value for it to resolve against.

Otherwise, you're throwing away core spec features (SVGLength is a multi-type string, not a number), and making it harder to replace your code with a correct version in future.

Source/DOM classes/Unported or Partial DOM/SVGSVGElement.m
```diff
@@ -133,7 +133,17 @@ We would need to put extra (NON STANDARD) properties on SVGDocument, for the "vi
 	NSString* stringWidth = [self getAttribute:@"width"];
 	NSString* stringHeight = [self getAttribute:@"height"];
 	
-	if( stringWidth == nil || stringWidth.length < 1 )
+    /**
+     Ignore percetage width and heights which are only used when rendering in HTML
```
adamgit, line 137: AFAICS these are not "only used when rendering in HTML" - they are potentially used in _all_ SVG rendering, specifically in any situation where the document is embedded in another SVG, or in a larger XML document,  or in a predefined viewport area (e.g. the iPhone's full screen)

adamgit:
Going through this in detail, I like what you're trying to do, but I think there's a cleaner way.
1. Add a method to SVGLength: "-(float) convertToPixelsUsingParentLength:(SVGLength) parent;". On absolute SVGLengths, this returns the existing 'convert-to-px'; on relative SVGLengths, it calculates the px of the parent SVGLength, and then multiplies by its relative amount.
2. -- EDIT: actually, better method is one that takes an SVGElement as argument. That way we can isolate the "specific to SVGSVGElement" code from 3. below into this one method above. --
3. Wherever we read the ".width" property of an SVGElement, we currently always assume it is absolute, and we crash / treat as 0 otherwise (which is incorrect!). Instead, we use the above method, and pass-in the SVGLength: "element.parent.width". In the special case of an SVGSVGElement, we override this and pass in "new SVGLength( element.viewBox.width )" (which might also be 0, but that's OK - in that case, we have no other options anyway!)

This way, we fix "relative" width / height everywhere, AND we keep the information about original viewport (which your patch currently discards), AND - I believe - the example SVG's you have will all be fixed.

...AND, as a bonus, other percentages for width/hegith on the root SVG will all "just work".

chetbox:
Looking at this in more detail there are two problems here:
- If `<svg>` specifies `width="X%"` and/or `height="Y%" we see a crash because we're trying to convert to a pixel value which isn't possible.
- If the root `<svg>` element does not specify a `width` or `height` attribute the SVG still renders correctly _until you try to scale the image_. i.e. After `[svgkImage scaleToFitInside:rect]` the image renders as a horizontal line.

It is trivial to create test cases:
- For the former case above: Edit an existing SVG and set the `width` and `height` attributes of `<svg>` to `100%`.
- For the latter: Edit an existing SVG to remove the `width` and `height` attributes of `<svg>`. (This loads correctly.) Now Add a `[loadedImage scaleToFitInside:CGRectMake(200, 200)]` in `DetailViewCotroller.m` and it now renders incorrectly.

Both of these cases are perfectly valid SVG files as far as popular editors are concerned.

I've been trying to work through this and I can't quite work out how to structure it. Your instructions seem clear enough, but there are a few things that don't quite make sense to me.
- `SVGSVGElement.width` and `SVGSVGElement.height` are `SVGLength`s. We need something that attempts to convert a relative `SVGLength` into a pixel `SVGLength`. I can't see how to create an absolute `SVGLength` without throwing away percentage information as you have suggested. Your suggestion of "new SVGLength( element.viewBox.width )" is what my `svgLengthFromNumber` is intended for. (There is no existing constructor to create a `SVGLength` from anything but a `NSString`.)
- You second point above doesn't quite work unless I make `width` and `height` public on `SVGSVGElement`. 

I believe all we need to do is:
- If a percentage width or height has been supplied and there is no concrete parent unit type, ignore it. (Either when rendering or when parsing.)
- If no absolute width or height is specified we must set it using the `viewBox` to ensure that scaling works correctly.

Any suggestions of how to proceed?

adamgit:
I think therrs a lot more than two problems. One fix can fix those AND several others.

Eg i believe we crash anywhere there is a % width or height.

Eg when svg width is a percentage, there are several cases where we should be calculating actual numbers - using the viewbox is actively wrong.

I'll have a look again later today, but I think we can retain percentage info easily enough.

adamgit:
I've made a personal branch / fork to play around with this:

https://github.com/adamgit/SVGKit/tree/FixPercentLengths

I've made a commit that does:
1. Adds a draft version of the "convert any percentage-SVGLength into the corect absolute/pixels SVGLenght" (in SVGHelperUtilities)
2. Alters SVGSVGElement so that instead of naively converting to pixels, and failing if it hits a percentage, it uses the above method to generate the "correct" absolute pixels
3. Deleted some code at end of SVGSVGElement method that shouldn't have been there
4. Added a method to convert SVGLength units into CSS primitive units (needed to make the first method above be correct)

As far as I can tell, this solves most of the above problem. What's missing:
1. in "SVGLengthFromLengthInSVGTree", replace:
   1. NSAssert(FALSE, @"Intelligent code goes here... (do things like use the ViewBox to guess a width/height)");"
2. ...with...
   1. The code proposed earlier in this thread
3. AND: look at the CALayer generation code, and make sure that wherever it needs "pixel value" of width or height, it uses the new SVGLengthFromLengthInSVGTree method instead, to guarantee that percentage lengths are handled

It may also be that we need to edit the "transform" code - if Percentage SVGlengths are legal inside a transform, then we should use SVGLengthFromLengthInSVGTree there too.

adamgit:
(in case it's not clear - I'm suggesting you fork that fork, see if you can easily apply the fixes on top.

If so, you can PR to my fork, and I can then merge and PR back into SVGKit master.

Or ditto if anyone else wants to fix this)

Because: this is quite a complex fix, impacting a bunch of files, and I'm not sure we'll get it right first time. I don't want to mess around too much in SVGKit upstream with this.

PR #240, opened 2014-11-15
Only apply clipping masks to elements that have the clip-path attribute.

dgileadi:
Suppose you have an SVG document like:

```
<g clip-path="url(#someClipPath)">
    <!-- lots of children here -->
</g>
```

The code would redundantly apply the clipping path from someClipPath to the layer of every one of those children, as well as to the &lt;g&gt; layer. The result was huge slowness if there were many children. This pull request fixes it to only apply the clipping path to the layer with the clip-path attribute. Child layers will get clipped by their parent layer.


adamgit:
If I understand correctly, before this change you could clone a layer in the middle and it would render correctly ... After this change, cloned layers will be corrupt?

If so, that's potentially a major step backwards. I know I've used cloning of sublayers a lot on several apps.

Perhaps a flag on svgkinageview classes to let then optionally override this in order to achieve higher performance when rendering?

Or an optional flag on the calayertree method in svgjinage to switch between "performance improved" and "100% correct" output?

dgileadi:
I’m not sure I understand what you’re asking.

Before this change, as SVGKImage created layers for each element in the SVG it would check whether this element _or any of its ancestors_ has a clip-path attribute, and if so it would create a new mask layer, render the clipPath’s content to the mask layer, and attach the mask layer to the element’s layer.

So in the example I gave, the <g> element and all of its children would create two layers each: the element’s layer and an attached mask layer. Each mask layer would duplicate the same content, and when drawing Apple would clip each sublayer to its own mask, and then to the parent layer's mask. If you were to clone one of the sublayers then its mask layer would also be cloned, assuming you use something like CALayer+RecursiveClone.

After the change the mask layer is only created for the <g> element’s layer; none of the sublayers would have its own mask layer and instead the sublayers are only clipped to their parent layer’s mask. If you cloned one of the sublayers then it wouldn’t have its own mask layer. However I wouldn’t consider this to be a “corrupt” sublayer; it would be perfectly usable. It would just have a nil `mask` property and would not be clipped to any mask. I’d actually tend to consider this to be correct behavior.

dgileadi:
So after I thought about it I figured I could come up with a sample SVG file that shows that the behavior in this patch is correct and the current behavior is incorrect. It was harder than I thought, and in the process I uncovered a bug with clip-path rendering on `<g>` elements. So this pull request now contains both the sample and a fix for that bug. Details follow on each:

### The Sample SVG

First, the sample: I created parent-clip.svg, which has the following in it (edited a bit for brevity):

``` SVG
    <rect x="100" y="100" width="100" height="100" fill="red" clip-path="url(#circular)" />
    <g clip-path="url(#circular)">
      <rect id="greensquare" x="100" y="100" width="100" height="100" fill="green" />
    </g>
```

The first `<rect>` and the `<g>` are clipped to the same circle. The actual file puts a couple `<animate>` elements inside the "greensquare" rect to move it around. If you open it in a browser you'll see the green rectangle move back and forth, always clipped to the circle thanks to the clip-path on its parent `<g>` element. If you open the parent-clip sample in the demo app with the changes in this pull request applied and tap the Animate button you'll see the same thing.

Without the original change in this pull request you'll see something else: a quarter circle moves back and forth. To see this apply the patch and then undo the code shown below. This is because the original code is incorrectly applying the mask from the parent `<g>` to both the parent and to its child `<rect>` as well. As a refresher the original change is in SVGKImage.m, and it changes:

``` Objective-C
    NSString* clipPath = [element cascadedValueForStylableProperty:@"clip-path"];
```

to:

``` Objective-C
    NSString* clipPath = [element getAttribute:@"clip-path"];
```

In other words, it makes the clip-path attribute not be inherited by child elements. This change agrees with the SVG spec, which says [the clip-path property is not inherited](http://www.w3.org/TR/SVG/masking.html#ClipPathProperty).

I hope this sample and the above explanation are make this patch clearer than it was before: the change is not only more performant but makes SVGKit match the spec, where the current behavior doesn't.

### The Bug Fix

As I noted at the start, while I was working on the sample SVG file I ran across a bug in my original clip-path implementation: when clipping a `<g>` element that contains children with origins that aren't at 0/0 it would clip the `<g>` element at the origin of the children. For example, in parent-clip.svg the red square would be correctly clipped to show a quarter-circle with the point at the top left, but the green square would be incorrectly clipped to show a quarter circle with the point at the bottom right.

This was happening because the mask frame is calculated before the SVGGElement has a chance to lay out its children and adjust its own frame to fit them. The fix is simple: I added code in SVGGElement where after it sets its own frame it checks whether it's masked, and if so it adjusts its mask's frame to match its new origin.

adamgit:
"Cascaded value ..."

Doesn't only mean "inherit", it also means " execute a full CSS cascade".

So .. I see your point that clippath shouldn't be inherited, but surely.you 
still very much need the CSS cascade to.run?

dgileadi:
You're right. I've added a cascadedValueForStylableProperty overload that lets you specify whether the property is inherited. I've also run through [the SVG spec's styling properties list](http://www.w3.org/TR/SVG/styling.html#SVGStylingProperties) and converted a few from our codebase that weren't handled correctly (opacity, stop-color and stop-opacity) to use the new method.

adamgit:
The new cascadedValueForStylableProperty:inherit: looks great. Much easier to see/understand the inheritance rules like this (especially opacity!)

dgileadi:
Agreed; I’m glad you pushed me towards it. I was a bit nervous changing opacity, so afterwards I took the SVG spec’s opacity01 example and put it in the demo app. I was very happy to see that it renders correctly.

PR #250, opened 2014-12-04
hack around renderInContext bug

iliaskarim:
`Apparently, they introduced a bug on the renderInContext when the view contains 0 in the width` - https://github.com/nicklockwood/FXBlurView/issues/79 (crashes)


adamgit:
c.f. #278 comments - the const change is probably wrong.

If you can provide a pull request that ONLY makes the changes you describe - great! I'll merge it.

But currently, this PR contains many other unrelated changes, most of which are not good. e.g. I don't like the change from int to long - should be a change to appropriate NS\* integer if anything, now that apple has redefined all integers.

iliaskarim:
hey @adamgit thanks for your response! I really only meant to draw your attention to the one commit  in question: [3928df9](https://github.com/djz-code/SVGKit/commit/3928df9beb097e7802c74e22f1bc620d99b0e233)

does that one look good to you? If so I'll prepare a pull request

adamgit:
That one looks OK to me.

Greg has been checking in some improvements to the "bounds" generation recently - might not be needed any more?

(I think it is needed, just saying: check with latest)

iliaskarim:
can confirm that the issue seems resolved in 1.2 👍🏻

PR #251, opened 2014-12-10
Lumberjack 2.0.0-beta4 migration

seltzered:
Updated to use cocoalumberjack 2.0.0-beta4. See https://github.com/seltzered/SVGKit/commit/b7808081b1766fc0a1e8b81e936f072a0d5e69b0 for a detailed note on how I added this via git subtree.

I was a bit torn between whether to do it this way or pullreq the method done via just adding a podfile -- this seemed lower risk for now and for those not using cocoapods. I do have an experiment via the podfile method in https://github.com/seltzered/svgkit/tree/Lumberjack-2.x-svgkit-experimental


adamgit:
Lumberjack 2 is now out of beta.

How can we best upgrade to 2.0 official?

seltzered:
Hey @adamgit , thanks for bringing this up - I'll look into it later this week. IIRC, @MaddTheSane made a better call when I pullreq'ed to his fork and installed Lumberjack 2.x via submodules instead of the git tree method I initially did. 

So basically I need to:
- switch the references to submodules (lemme know if there's any reason I shouldn't)
- see if there's any extra fixes me or @MaddTheSane did back in December -- particularly on that fork there were some small changes to the podfile.

Currently have been using SVGKit+Lumberjack 2.x personally by pointing cocoapods to MaddTheSane's fork, things have been working fine so far.

adamgit:
We've had recurring problems with submodules - each newcomer has to learn 
how to use them with their particular git client. And when they get it 
wrong, your repository is screwed (git can easily get to corrupted state 
with submodules, where standard fixes like git checkout simply stop working 
:( ).

We need a solution that doesn't rely upon submodules :).

seltzered:
Hum, so would sticking with git read-tree and leaving instructions on how to update Lumberjack in Readme.md (or alternatively in a readme in the Vendor folder) suffice then?

i.e. the instructions would be similar to the following from one of my earlier commit messages:
- git remote add CocoaLumberjack_remote https://github.com/CocoaLumberjack/CocoaLumberjack.git
- git fetch CocoaLumberjack_remote
- git checkout -b CocoaLumberjack CocoaLumberjack_remote/master
- switch back to the branch of SVGKit you're working on
- git read-tree --prefix=Source/Vendor/CocoaLumberjack/ -u 2.0.0-beta4:Classes (the tag, which may conflict with other tags, can alternatively be a commit hash instead)

adamgit:
I'd like to see lumberjack 2.0.x merged directly into our 2.x branch (i.e. the main active one), and no-one has to do anything manual. We'll then support only one version of lumberjack (whichever one is currently merged).

Ideally ... I'd like to see lumnberjack built / embedded as a Framework, so that users can selectively include/remove/replace it when building their app projects. AIUI, if the framework isn't present when the user's app is built, it'll fail (because frameworks are linked not embedded). So ... user would either "embed the copy of framework we bundled in svgkit" or "embed whatever different version of lumberjack they prefer, as a framework".

?

seltzered:
Ah, okay. Looking into this now.

seltzered:
Added as a framework. Looks like it works, though I've only ran the demo in the simulator (don't have an iOS device atm).

adamgit:
There's a lot of crap in there from Lumberjack that shouldn't be - github refuses to even show the diff :(. Is it possible to remove things like the Benchmarking folders, and the Demos, while still including this as a framework?

Should be a verion of the framework that ONLY builds the library, and nothing else (like we have with SVGKit)

seltzered:
Okay, stripped things down to only build the framework.

adamgit:
That looks a lot better.

Is it possible to also remove the "LibTest", "SwiftTest", "iOSSwift" and "FmwkTest" folders, or are they integral to the framework?

(they look like unit tests or something, which we should be omitting if we can - but I don't know how this is packaged, so maybe they contain something important? PRevious Lumberjack was very "light" as library, only needed to grab the .h and .m files)

adamgit:
Also, once this is in, I think it's a good time to do a new point release of SVGKit.

We'll have:
- 2.x Lumberjack
- New Demo app
- Lots of small rendering fixes

...although I'm hoping for a few more fixes too from @greghe wrt gradients.

adamgit:
(all we'd need now is a rewrite of the Text handling, and it would be great :))

seltzered:
Okay, removed the dirs mentioned, things still seem to work. 

Leaving in the OSX Desktop framework build project in though, would like to someday talk about how to help in merging @MaddTheSane 's fork.

adamgit:
I'm going ahead and merging this ... let's see how well it works in the 2.x branch for a while.

I'll hold off on accepting other pull requests for a bit ...if anyone reports problems, I'll back it out ... but this is a pill we need to swallow :), so let's get started...

adamgit:
Re: the fork, for the bits shared across iOS and OS X we need to break it
down into single-feature commits that we can review, merge, one by one.

PR #253, opened 2014-12-16
Fix in SVGKit wasting lot of memory in CALayer

krzysiek84:


adamgit:
To confirm: this has no bad effects on eg hit testing the layer?

krzysiek84:
Can not confirm, as i did test only on svgs provided in our app. And don't think i use any hit testing.
Well don't think it should affect hit testing, the way (i think) SVG Kit uses layer is only to provide the renderInContext. So proper fix would be to reimplement it so that each element is not the CALayerSubclass. For me just making that display do nothing helps and i'm able to run app on iPad 1.

krzysiek84:
Ah and as mentioned in comment, The bug is visible when the svg has large area set in xml.
Guess if it would be more than 2048x2048 (max layer size on old ipad) it would crash instantly.
Not sure if that bug is visible at all when no size is specified.

adamgit:
Can you do a quick test with the supplied svgkit-demo project?

If you load any svg - eg the monkey - and tap anywhere, it will hottest and 
draw a box around the sublayer hit first. Can press the debug button in top 
right to get coloured outline etc

krzysiek84:
Running on iPhone 6 simulator it seams to work ok, i can see green area around eyes or ears :smiley: 

So bug is : if i have svg with size defined as 1000x1000, even if i render the image at 10x10, it still uses 1000x1000 for the layer, because CALayerWithClipRender calls new_shmem when renderInContext is called.

adamgit:
I want to merge this, but it seems too too to be true.

Presumably there's a downside to disbaljng the backing layer. Maybe: re 
rendering an Svgklayeredimageview is much slower?

krzysiek84:
I fully understand it's risky change. Moreover SVG Kit might be broken for users who use the SVGImage.CALayerTree directly (but not sure if anyone uses it this way).

For sure there is a bug with memory being eaten by 'invisible' layers. That is layer of size 1000x1000 creates back buffer image of same size, even though it's rendered as 10x10 block on screen.

As i see it the CALayerWithClipRender is used not as layer (added to the view) but rather as 'rendering recipe' - only renderInContext: in fast image view. My fix would most likely not work if the layer was added to the view layer..  
And i use only SVGKFastImageView which just calls renderInContext so all is good. Rendering is not slower as the FastImageView layer is backed (but this time as 10x10 size not 1000x1000).

But anyway, question is how to draw the svg path elements without backing it in memory. My first thought was to rewrite it so that each element does not use CALayer, but as there is hit testing etc involved that might take time. So easier would be to use layers but disable backing them in memory. And looks like the hack with empty display does the trick (tested on iphone 5s, ipad 1 and simulators).

adamgit:
Many people use CALayerTree - it's one of the core features.

I'm pretty sure we can't merge this as-is. Although it saves memory in your
case, in many cases it will massively reduce performance - the CPU cost of
re-calculating e.g. CGPath is very high, too high to do in realtime (I've
experimented with this in the past - Apple's drawing libraries are very
slow!).

From the docs, it looks like you could achieve the same effect at runtime
by assigning a custom CALayer.delegate:

" The layer calls this method at appropriate times to update the layer’s
content. If the layer has a delegate object, this method attempts to call
the delegate’s displayLayer: method, which the delegate can use to update
the layer’s contents. If the delegate does not implement the displayLayer:
method, this method creates a backing store and calls the layer’s
drawInContext:
https://developer.apple.com/library/mac/documentation/GraphicsImaging/Reference/CALayer_class/index.html#//apple_ref/occ/instm/CALayer/drawInContext:
method to fill that backing store with content. The new backing store
replaces the previous contents of the layer."

In that case, you could do:
1. add a method to SVGKFastImageView, something like "-(void)
   removeBackingStores", which recurses down the CALayerTree, calling
   "delegate = BLAH" , where BLAH is a new class that has an approppriate
   (empty?) displayLayer: method.

..this achieves the same effect, but is:
1. Optional
2. Disabled by default
3. Only applied to FastImageView (AFAICS, if you're using LayeredImageView,
   you'd probably never want to use this approach?)

?

PS ... I suspect that your change reduces performance for all graphics? My
suspicion is that the "backing store" is implemented as OpenGL Texture RAM.
If so: removing the backing store prevents Apple from caching layers
internally on the GPU.

krzysiek84:
No performance is not an issue at all, it draws it once to fastimageview layer. so once it's drawn all the panning and stuff works nicely. 

Moreover i do have custom view too which calls:

```
CGContextSaveGState(context); {
            CGContextTranslateCTM(context, offset.x, offset.y);
            CGContextScaleCTM(context, scale, scale);

            [svgIcon.CALayerTree renderInContext:context];
        }
        CGContextRestoreGState(context);
```

And this also is not affected by the change (as i don't expect CALayerTree to be back buffered anyway, scale is changed dynamicaly).

Current implementation (CALayer subclasses with backing on) wastes a lot of memory.
This could be further investigated. 
Simple fix would be to discourage users from using CALayerTree from SVGImage as layer added to UIView and just allow using the renderInContext: to be used. That way my simple fix could be used. 
Other idea which comes to my mind is to rewirte the rendering to not use layers at all (except the part where the svg is rendered to view). 

Test for that would be - having svg with some ridiculous size (10k x 10k). I guess the lib will crash when trying to render svg on small area. As it has limit on texture size.

krzysiek84:
Did not noticed LayeredImageView before, think other way to fix it would be to call inside SVGKLayer renderLayerInContext instead of adding new layer to it.

adamgit:
Note: renderInContext probably doesn't fully work (officially, according to
Apple)

There are two reasons we have two versions of SVGKImageView:
1. a lot of people need access to all the sublayers (e.g. for hit-testing)
2. if your SVG's use: Gradients, drop-shadows, clipping, masking - or
   anything that uses those features behind the scenes - renderinContext will
   corrupt the render. Apple's response "not implemented yet"

(there are reports that they MIGHT have fixed some/all of this for iOS8,
but it was definitely still broken in iOS7)

krzysiek84:
If getting rid of layers is not an option than disabling the buffering on any layer except for root layers of SVGKImageView should work.

Using the lib as it is with big svgs displayed on small area is memory overkill.
And to be honest it might slow down the rendering too, as it takes time to generate the backing images as well. 
Think it was twice faster to render the grid of SVGFastImageViews (about 100) in a scroll view with the fix than without.

adamgit:
If we keep this de-buffering optional (there are people who are grabbing
the sub-layers frequently - they'd want to leave "buffer everything" turned
on, I think?), that sounds to me like a great solution.

As you describe, IMHO it could be made "default disabled (no buffering
except on root layer)".

krzysiek84:
Yeah but problem with setting is that most people will not go too much detail on project config, they will get the lib, create svgimageview and add it somewhere :D 
Best would be to set optional so that it fixes issue, but add warning/exception, if they try to use layer with the hack as 'standard layer'. Not sure really how to achive that. 
But still optional with on (buffer everything) is better than nothing. Even some info about the bug in comments would help i suppose. Maybe someone can get some better fix.

adamgit:
I think it's enough that we put class-level code comment on e.g.
SVGKLayeredImageView to say "this is de-buffered by default, because most
cases this saves you memory. But if you are accessing the layers directly,
this will cost you big performance."

And copy/paste teh same comment to the [SVGKImage CALayerTree] method.

...because, I think, those are the ONLY two places someone would be likely
to look if they're going to do layer-by-layer fetching?

ULTIMATELY: the clever thing to do would be to include a heuristic where we
check the size of image versus render size, and output a log WARNING if we
think you should have buffering off (but it's on), and vice versa :)

krzysiek84:
I like the ultimately solution :D 
Logging some warning if user uses the layer wrong way sounds like a nice way to let the clients know that api has changed a bit. 
How do you think the layered image view should be changed so that it does not use the buffering, making svgklayer to call renderInContext would not break stuff ?

PR #256, opened 2014-12-18
Improve parsing to avoid redundant copying and stack allocation

dgileadi:
I was getting an error with an SVG image that has a large embedded image inside it (using an xlink:href attribute with a `data:image/png;base64` value)—the error was happening in SVGKParser when trying to strncpy the xlink:href attribute. I realized that there should be no need to strncpy the data; instead NSString's initWithBytes:length:encoding can be used directly. I changed the code to do so, which should make it more efficient and also avoids the error I was getting.

### DIFF
--- a/Source/Parsers/SVGKParser.m
+++ b/Source/Parsers/SVGKParser.m
@@ -715,11 +715,7 @@ static void	endElementSAX (void *ctx, const xmlChar *localname, const xmlChar *p
 }
 
 - (void)handleFoundCharacters:(const xmlChar *)chars length:(int)len {
-		char value[len + 1];
-		strncpy(value, (const char *) chars, len);
-		value[len] = '\0';
-		
-		[_storedChars appendString:[NSString stringWithUTF8String:value]];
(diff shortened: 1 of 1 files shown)


Source/Parsers/SVGKParser.m
```diff
@@ -873,16 +869,13 @@ static void structuredError		(void * userData,
 	for (int i = 0; i < attr_ct * 5; i += 5) {
 		const char *begin = (const char *) attrs[i + 3];
 		const char *end = (const char *) attrs[i + 4];
-		size_t vlen = strlen(begin) - strlen(end);
```
adamgit, line 876: If this used to work (begin-end) does your version still work (end-begun)

dgileadi, line 876: Yes. The old version was doing strlen(begin) - strlen(end), and strlen() counts from the pointer until it finds a zero (‘\0’) terminator. This worked because the zero terminator was always further from the begin pointer than from the end pointer, so strlen(begin) was always bigger than strlen(end). However, there’s no need to count string lengths at all—subtracting the begin pointer from the end pointer gives the same length with no counting required.

PR #263, opened 2015-01-27
Fixed crash when changing a SVGKLayeredImageView image

buscarini:
If I changed a SVGKLayeredImageView image property it would crash on _SVGKImage release because the property observers hadn't been removed. This fixes the issue. Maybe a try catch is not the best way of doing this, but it works.


buscarini:
I have also fixed another issue with iOS 8 and empty layers when rendering to a context

Svantulden:
Thanks for this! I had this problem too for the `SVGKFastImageView`.

PR #268, opened 2015-02-03
Not import unnecessary DOM headers to reduce references.

rhcad:

PR #278, opened 2015-02-27
Not import unnecessary DOM headers to reduce references.

rhcad:
Re-submit #268 based on the newest 2.x branch.

### DIFF
--- a/Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
@@ -7,6 +7,7 @@
 #import "SVGKImage.h"
 #import "SVGKSourceURL.h"
 #import "SVGKSourceNSData.h"
+#import "SVGSVGElement.h"
 
 #if TARGET_OS_IPHONE
 
--- a/Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
@@ -106,7 +106,7 @@ CGImageRef SVGImageCGImage(AppleNativeImageRef img)
 	 */
 	AppleNativeImageRef image = [AppleNativeImage imageWithData:imageData];
 	
-    if( image == nil ) // NSData doesn't contain an imageformat Apple supports; might be an SVG instead
+    if( image == nil && imageURL ) // NSData doesn't contain an imageformat Apple supports; might be an SVG instead
     {
         SVGKImage *svg = nil;
         
--- a/Source/QuartzCore additions/CGPathAdditions.m
+++ b/Source/QuartzCore additions/CGPathAdditions.m
@@ -21,7 +21,7 @@ CGFloat fixInfinity(CGFloat inputFloat){
     return inputFloat;
 }
 
-CGPoint *fixPointsInfinity(CGPathElement *element){
+CGPoint *fixPointsInfinity(const CGPathElement *element){
     int i,total;
     
     switch (element->type) {
--- a/Source/SVGKImage.h
+++ b/Source/SVGKImage.h
@@ -111,6 +111,7 @@ typedef void (^SVGKImageAsynchronousLoadingDelegate)(SVGKImage* loadedImage);
 + (SVGKImage*) imageWithSource:(SVGKSource *)newSource; // if you have custom source's you want to use
 
 - (id)initWithContentsOfFile:(NSString *)path;
+- (id)initWithData:(NSData *)data;
 
 #pragma mark - UIImage methods cloned and re-implemented as SVG intelligent methods
 
--- a/Source/SVGKImage.h
+++ b/Source/SVGKImage.h
@@ -111,7 +111,6 @@ typedef void (^SVGKImageAsynchronousLoadingDelegate)(SVGKImage* loadedImage);
 + (SVGKImage*) imageWithSource:(SVGKSource *)newSource; // if you have custom source's you want to use
 
 - (id)initWithContentsOfFile:(NSString *)path;
-- (id)initWithData:(NSData *)data;
 
 #pragma mark - UIImage methods cloned and re-implemented as SVG intelligent methods
 
--- a/Source/DOM classes/Core DOM/Node.m
+++ b/Source/DOM classes/Core DOM/Node.m
@@ -200,7 +200,7 @@
     if( [self.childNodes length] < 1 )
         return nil;
     else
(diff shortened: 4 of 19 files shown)


Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
```diff
@@ -105,7 +106,7 @@ - (CALayer *) newLayer
 	 */
 	AppleNativeImageRef image = [AppleNativeImage imageWithData:imageData];
 	
-    if( image == nil ) // NSData doesn't contain an imageformat Apple supports; might be an SVG instead
+    if( image == nil && imageURL ) // NSData doesn't contain an imageformat Apple supports; might be an SVG instead
```
adamgit, line 109: This is incorrect. You do't need an imageURL for this if statement.

Source/QuartzCore additions/CGPathAdditions.m
```diff
@@ -21,7 +21,7 @@ CGFloat fixInfinity(CGFloat inputFloat){
     return inputFloat;
 }
 
-CGPoint *fixPointsInfinity(CGPathElement *element){
+CGPoint *fixPointsInfinity(const CGPathElement *element){
```
adamgit, line 24: I don't understand this. We change the element - surely this cannot be const?

Source/SVGKImage.h
```diff
@@ -111,7 +111,6 @@ typedef void (^SVGKImageAsynchronousLoadingDelegate)(SVGKImage* loadedImage);
 + (SVGKImage*) imageWithSource:(SVGKSource *)newSource; // if you have custom source's you want to use
 
 - (id)initWithContentsOfFile:(NSString *)path;
-- (id)initWithData:(NSData *)data;
```
adamgit, line 114: It's not implemented yet, but we do need this method. Please don't delete it.

If preferred, you can add an empty method body to the .m file, which NSAssert (false, @"Not implemented yet");

Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
rhcad, line 109: [SVGKSourceURL sourceFromURL:imageURL] will crash on iOS 8 if imageURL is nil, so I check imageURL at the line `if( image == nil )`.

Source/QuartzCore additions/CGPathAdditions.m
rhcad, line 24: There's a warning about passing parameter to fixPointsInfinity in applier and applyPathTranslation which has const element parameter:
`void applier (void *info, const CGPathElement *element)`
`void applyPathTranslation (void *info, const CGPathElement *element)`

Source/SVGKImage.h
rhcad, line 114: I can add the empty method body as your comment. Thanks.
By the way, how to get the data of SVGKImage object? And I think we do not need cache it which come from SVG file or has large content.

Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
adamgit, line 109: OK, but that bug needs to be fixed. Breaking this for all versions of iOS isn't workable!

Do we have a bug report (even better: a workaround/fix?) for the ios8 crash? Is it simply this line that's causing it:

NSInputStream\* stream = [NSInputStream inputStreamWithURL:u];

Source/QuartzCore additions/CGPathAdditions.m
adamgit, line 24: I see what you mean.

I think the correct fix is to change line 98 from:

void applyPathTranslation (void *info, const CGPathElement *element) {

to:

void applyPathTranslation (void *info, CGPathElement *element) {

?

rhcad:
I have found two reasons of the crash at `[self removeObserver:self forKeyPath:...` where may has `SOMETIMES CRASHES IN APPLE CODE, CAN'T WORK OUT WHY` comment. 
1. Can not remove the observer which has't added.
2. The observer should be removed before destroyed (in dealloc function).
We should check whether the observer has added, such as `if (_showBorder) { [self removeObserver:self forKeyPath:@"showBorder"];}`.

Source/QuartzCore additions/CGPathAdditions.m
rhcad, line 24: I don't agree with it for keeping const validation and `void applier (void *info, const CGPathElement *element)` can not be changed to `..., CGPathElement *element)`.

Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
rhcad, line 109: You may be right, but it's not necessary to continue if imageURL is nil due to parse fail. I'm not against the two methods you mentioned.

Source/QuartzCore additions/CGPathAdditions.m
adamgit, line 24: Code cannot work with const, though.

I think ... "CGPoint *fixPointsInfinity(CGPathElement *element)" needs to be rewritten to:
1. allocate a new array
2. put changed values in new array, leaving old array as const
3. renamed to make it clear it allocates a new array

And "applyPathTranslation" has to be changed so that:
1. uses the new-named function
2. releases the new array at end of function

Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
adamgit, line 109: Instead of:

 "&& imageURL"

let's change:

"if( effectiveSource == nil )"
to
"if( effectiveSource == nil && imageURL != nil )"

...effectiveSource will remain nil, but there's still the possibility for somoene to fix the iOS8 bug in SVGKSourceURL later

Source/QuartzCore additions/CGPathAdditions.m
rhcad, line 24: Ok, you can use `CGPoint buf[4];` instead of allocating a new array.

Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
rhcad, line 109: OK. What come next?

adamgit:
Too much of this has changed - github won't let me merge it.

Also, looking through the changes, some of them I'd have to reject anyway.

e.g. you should not use "int" in iOS any more - NSUInteger/NSInteger instead (part of the required changes Apple pushed for 64 bit compatibility).

PR #288, opened 2015-03-25
Add support for skewX and skewY transformations

greghe:


adamgit:
This takes out the error-reporting, but doesn't add any code.

I think you submitted a PR for the wrong commit

greghe:
There are 8 lines of new code. I followed the existing style very closely, so perhaps they don't look new?

adamgit:
Ah - bug in GitHub! :(

It's only showing the deletions, although if you go to the source file, the additions are there too.

Sorry.

PR #291, opened 2015-03-28
More robust inline style parsing.

greghe:
Inline style declarations containing spaces were being ignored. With this change, a style declaration like:

```
... style="fill: orange; stroke : black ; stroke-width: 2"/>
```

is now parsed correctly.

### DIFF
--- a/Source/DOM classes/Core DOM/CSSStyleDeclaration.m
+++ b/Source/DOM classes/Core DOM/CSSStyleDeclaration.m
@@ -74,7 +74,7 @@
         /* SVG spec specifies some attribute lists might be delimited by white space.
          e.g. stroke-dasharray is 'A list of comma and/or white space separated <length>s'
          http://www.w3.org/TR/SVG/painting.html */
-		if (c == '\n' || c == '\t') {
+		if (isspace(c)) {
 			continue;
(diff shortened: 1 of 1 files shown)


Source/DOM classes/Core DOM/CSSStyleDeclaration.m
```diff
@@ -74,7 +74,7 @@ -(NSMutableDictionary *) NSDictionaryFromCSSAttributes: (NSString *)css {
         /* SVG spec specifies some attribute lists might be delimited by white space.
          e.g. stroke-dasharray is 'A list of comma and/or white space separated <length>s'
          http://www.w3.org/TR/SVG/painting.html */
-		if (c == '\n' || c == '\t') {
+		if (isspace(c)) {
```
adamgit, line 77: where does isspace come from? I can't find it in the project nor in Apple docs?

MaddTheSane, line 77: `isspace()` is a C function that determines if a character is of a few selected types. For more information, there is a man page with more info. Just type `man isspace` in a Terminal window.
And yes, `\n` and `\t` are covered.

PR #298, opened 2015-03-29
Add support for relative coordinates and lengths to circles, ellipses and lines.

greghe:
Also fixes an issue where perfectly horizontal or vertical lines were not being rendered (empty bounding rectangle).

### DIFF
--- a/Source/DOM classes/SVG-DOM/SVGHelperUtilities.m
+++ b/Source/DOM classes/SVG-DOM/SVGHelperUtilities.m
@@ -318,20 +318,37 @@
 	
 	[self configureCALayer:_shapeLayer usingElement:svgElement];
 	
+	NSString* actualStroke = [svgElement cascadedValueForStylableProperty:@"stroke"];
+	if (!actualStroke)
+		actualStroke = @"none";
+	NSString* actualStrokeWidth = [svgElement cascadedValueForStylableProperty:@"stroke-width"];
+	
+	CGFloat strokeWidth = 1.0;
+	
+	if (actualStrokeWidth)
+	{
+		SVGRect r = ((SVGSVGElement*) svgElement.viewportElement).viewport;
+		
+		strokeWidth = [[SVGLength svgLengthFromNSString:actualStrokeWidth]
+					   pixelsValueWithDimension: hypot(r.width, r.height)];
+	}
+	
 	/** transform our LOCAL path into ABSOLUTE space */
 	CGAffineTransform transformAbsolute = [self transformAbsoluteIncludingViewportForTransformableOrViewportEstablishingElement:svgElement];
-	CGMutablePathRef pathToPlaceInLayer = CGPathCreateMutable();
-	CGPathAddPath( pathToPlaceInLayer, &transformAbsolute, pathRelative);
+
+	// calculate the rendered dimensions of the path
+	CGRect r = CGRectInset(CGPathGetBoundingBox(pathRelative), -strokeWidth/2., -strokeWidth/2.);
+	CGRect transformedPathBB = CGRectApplyAffineTransform(r, transformAbsolute);
+	
+	CGPathRef pathToPlaceInLayer = CGPathCreateCopyByTransformingPath(pathRelative, &transformAbsolute);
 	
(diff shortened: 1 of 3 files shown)


Source/DOM classes/Unported or Partial DOM/SVGLineElement.m
```diff
@@ -41,6 +52,10 @@ -(CALayer *)newLayer
 	CGPathAddLineToPoint(path, NULL, _x2, _y2);
 	
 	CALayer* result = [SVGHelperUtilities newCALayerForPathBasedSVGElement:self withPath:path];
+	if (CGRectIsEmpty(result.bounds)) // empty rects don't get rendered
```
adamgit: This is a bug in Apple's [CALayer bounds], right?

Because a CGPath of a horizontal or vertical line (and Apple knows this!) should not have an empty bounds. My understanding (Apple hasn't documented it properly): the bounds is meant to be the "rendered" bounds, i.e. including the fact that lines have 0.5 pixesl width (at least in CA they do: by default they are 1 pixel wide, with half the line rendered either side in pixels).

Presumably, we are also affected by the same bug in SVGRectElement, SVGPolylineElement, etx?

adamgit: Also: if a line is rendered from (0,0) to (0,0) ... what happens?

Appears to be a bug in the SVG spec: I can't find any statement on this :).

In the code you've got here, we'd generate a 1x1 bounding box, and potentiall render a width-wide pixel dot. Since the spec appears incomplete, I'd say that's OK?

greghe:
I will look into it.

Source/DOM classes/SVG-DOM/SVGHelperUtilities.m
```diff
@@ -318,20 +318,37 @@ +(CALayer *) newCALayerForPathBasedSVGElement:(SVGElement<SVGTransformable>*) sv
...
+	CGFloat strokeWidth = 1.0;
+	
+	if (actualStrokeWidth)
+	{
+		SVGRect r = ((SVGSVGElement*) svgElement.viewportElement).viewport;
+		
+		strokeWidth = [[SVGLength svgLengthFromNSString:actualStrokeWidth]
+					   pixelsValueWithDimension: hypot(r.width, r.height)];
+	}
+
```
greghe, line 335: I moved the stroke width calculation up so that it can be used to calculate the rendered bounding box.

Source/DOM classes/SVG-DOM/SVGHelperUtilities.m
```diff
@@ -318,20 +318,37 @@ +(CALayer *) newCALayerForPathBasedSVGElement:(SVGElement<SVGTransformable>*) sv
...
+	
 	/** transform our LOCAL path into ABSOLUTE space */
 	CGAffineTransform transformAbsolute = [self transformAbsoluteIncludingViewportForTransformableOrViewportEstablishingElement:svgElement];
-	CGMutablePathRef pathToPlaceInLayer = CGPathCreateMutable();
-	CGPathAddPath( pathToPlaceInLayer, &transformAbsolute, pathRelative);
+
+	// calculate the rendered dimensions of the path
+	CGRect r = CGRectInset(CGPathGetBoundingBox(pathRelative), -strokeWidth/2., -strokeWidth/2.);
+	CGRect transformedPathBB = CGRectApplyAffineTransform(r, transformAbsolute);
+
```
greghe, line 342: The bounding box is outset by 1/2 the line width before the absolute transformation is applied.

greghe:
It turns out that `CGPathGetBoundingBox()` is returning the geometric box not the rendered box. And since points and lines have no inherent width, returning an empty bounding box is possible. This also means that SVGKit is often drawing outside the bounds of a shape's CALayer. This is usually not a problem (since the drawing is not being clipped to the CALayer) - except in the case when the CALayer has a zero sized height or width, in which case it is not drawn at all.

This change now includes a path's stroke-width in the calculation of its bounding box - so now all path drawing should occur within the bounds its CALayer. And a zero-sized CALayer will be produced only when the stroke-width itself is zero.

PR #300, opened 2015-03-31
Improve linear gradients

greghe:
Makes several improvements to linear gradients:

xlink hrefs are now followed for gradients to acquire properties from other gradients
gradient strokes are now supported
better path masking for gradients

To do: improve linear gradients that are using `userSpaceOnUse` coordinates.

gradients

### DIFF
--- a/Source/DOM classes/SVG-DOM/SVGHelperUtilities.m
+++ b/Source/DOM classes/SVG-DOM/SVGHelperUtilities.m
@@ -314,7 +314,7 @@
 
 +(CALayer *) newCALayerForPathBasedSVGElement:(SVGElement<SVGTransformable>*) svgElement withPath:(CGPathRef) pathRelative
 {
-	CAShapeLayer* _shapeLayer = [[CAShapeLayerWithHitTest layer] retain];
+	CAShapeLayer* _shapeLayer = [CAShapeLayerWithHitTest layer];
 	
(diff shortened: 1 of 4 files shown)


Source/DOM classes/SVG-DOM/SVGHelperUtilities.m
```diff
@@ -423,76 +425,120 @@ +(CALayer *) newCALayerForPathBasedSVGElement:(SVGElement<SVGTransformable>*) sv
...
+				strokeLayer.lineJoin = kCALineJoinBevel;
 		}
 		if( actualMiterLimit.length > 0 )
 		{
-			_shapeLayer.miterLimit = [actualMiterLimit floatValue];
+			strokeLayer.miterLimit = [actualMiterLimit floatValue];
 		}
+		if ( [actualStroke hasPrefix:@"url"] )
+		{
+			// need a new fill layer because the stroke layer is becoming a mask
```
adamgit, line 449: Please duplicate / reference this comment on line 372. Otherwise, it looks like a bug / typo that you have two references to the same object. I can see what you're doing, seems good, but would help a lot to make this clear.

adamgit:
lingrad01 looks better, but this appears to have broken radial gradients.

c.f. radgrad01 (in specs section) and the sunburst image.

PR #301, opened 2015-04-01
Improvements for userSpaceOnUse linear gradients.

greghe:
The change here is to pass down the absolute affine transform, so that the userSpaceOnUse linear gradient can be properly transformed. Inkscape seems to favor this type of gradient, and, as shown below, our rendering is now more in line with Inkscape (and other SVG renderers) with these changes.
gradients2

### DIFF
--- a/Source/DOM classes/SVG-DOM/SVGHelperUtilities.m
+++ b/Source/DOM classes/SVG-DOM/SVGHelperUtilities.m
@@ -358,7 +358,6 @@
 	
 	_shapeLayer.path = finalPath;
 	CGPathRelease(finalPath);
-	CGPathRelease(pathToPlaceInLayer);
 	
 	/**
 	 NB: this line, by changing the FRAME of the layer, has the side effect of also changing the CGPATH's position in absolute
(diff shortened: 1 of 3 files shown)


Source/DOM classes/SVG-DOM/SVGHelperUtilities.m
```diff
@@ -490,15 +490,17 @@ +(CALayer *) newCALayerForPathBasedSVGElement:(SVGElement<SVGTransformable>*) sv
...
+		
 		CAShapeLayer* maskLayer = [CAShapeLayer layer];
 		maskLayer.frame = localRect;
 		maskLayer.path = fillLayer.path;
 		maskLayer.fillColor = [UIColor blackColor].CGColor;
 		maskLayer.strokeColor = nil;
 		gradientLayer.mask = maskLayer;
-		
+		if ( [gradientLayer.type isEqualToString:kExt_CAGradientLayerRadial])
+			gradientLayer.maskPath = pathToPlaceInLayer;
```
greghe, line 503: This should have been in my last check in, since i haven't worked on radial gradients.

adamgit, line 503: Tux looks a great deal better now (although eyes still missing), but
radgrad01 from the Spec is still worse than before the recent spate of
commits?

PR #303, opened 2015-04-04
Add support for the elliptical arc command in paths.

greghe:


adamgit:
Can you add this to the commit too?

http://www.w3.org/TR/SVG/images/paths/arcs01.svg

(new process for adding test files: https://github.com/SVGKit/SVGKit/wiki/Testing-SVGs -- NB: it looks long, but it's really quick. Unlike previous version, it's mostly automated now!)

greghe:
OK, I added both arcs files.

greghe:
There's an issue with arcs01.

PR #304, opened 2015-04-04
Ensure that elements referenced by <use> tags have the proper transform.

greghe:
Elements referenced by `<use>` elements now conform to the transform at their inserted location. The fix here is to save parent node of the inserted element, then set the parent node to the `<use>` element itself, and finally restore the original parent after rendering.

(This fix is needed for the "arcs02" example from the SVG standard to display correctly.)

### DIFF
--- a/Source/SVGKImage.m
+++ b/Source/SVGKImage.m
@@ -612,14 +612,21 @@ static NSMutableDictionary* globalSVGKImageCache;
 	//DEBUG: DDLogVerbose(@"[%@] DEBUG: converted SVG element (class:%@) to CALayer (class:%@ frame:%@ pointer:%@) for id = %@", [self class], NSStringFromClass([element class]), NSStringFromClass([layer class]), NSStringFromCGRect( layer.frame ), layer, element.identifier);
 	
 	NodeList* childNodes = element.childNodes;
-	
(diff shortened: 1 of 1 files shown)


adamgit:
Nice. Tux is now working 100%, and the simplified Reinel compass rose is ALMOST working (looks like something not quite right about the local rotations on the transforms of each use element).

adamgit:
Reinel is failing because:

"transform: scale(...)"

is transforming the object around its local center, rather than around the correct origin.

Reinel does: "scale(-1,1)", which should:
- mirror the object
- shift it left by the width of object
- ... so that it ends up back-to-back with original

But our "scale(-1,1)" currently does:
- mirror the object

...because it is using wrong origin

adamgit:
The Scale works correctly if there is no "use" element.

It appears to be something to do with use elements changing the local origin for Apple scale transforms.

adamgit:
I've pushed a trivial example - SVG with one Path, and one Use that clones the path:

Reinel_compass_rose-simplified-evenmore.svg

c.f. the preview image (should get a small arrow stacked on top of a large arrow). Instead, when rendering, we get a large arrow and the small arrow rendering underneath (Because its local origin was incorrect).

PR #305, opened 2015-04-04
More gradient improvements

greghe:
A few gradient improvements:
- radial gradients with `userSpaceOnUse` are now supported
- linear gradients now rotate when their bounding object rotates
- radial gradients are now clipped to their bounding box
- fixes an issue in which rotating a shape 90° could cause its stroke to disappear

### DIFF
--- a/Source/DOM classes/Unported or Partial DOM/SVGEllipseElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGEllipseElement.m
@@ -53,7 +53,7 @@
 	if( [[self getAttribute:@"r"] length] > 0 ) { // circle
 		
 		self.ry = self.rx = [[SVGLength svgLengthFromNSString:[self getAttribute:@"r"] ]
-							 pixelsValueWithDimension:hypot(r.width, r.height)];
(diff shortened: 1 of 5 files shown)


Source/DOM classes/Unported or Partial DOM/SVGEllipseElement.m
```diff
@@ -53,7 +53,7 @@ - (void)postProcessAttributesAddingErrorsTo:(SVGKParseResult *)parseResult {
 	if( [[self getAttribute:@"r"] length] > 0 ) { // circle
 		
 		self.ry = self.rx = [[SVGLength svgLengthFromNSString:[self getAttribute:@"r"] ]
-							 pixelsValueWithDimension:hypot(r.width, r.height)];
+							 pixelsValueWithDimension:hypot(r.width, r.height)/M_SQRT2];
 	}
```
greghe, line 57: Correcting formula.

adamgit:
Great!

Seems to have fixed everything except radialGradientTest.svg (but I didn't check that in last few commits, so I'm not sure which one broke it. I only know that it used to work)

PR #307, opened 2015-04-07
File seems to have an illegal move to command that causes an assert.

greghe:


adamgit:
Yes, but we can't change the file to make it work :). Inkscape, Apple, etc all render this file without complaint.

To be compliant with spec, we should still assert (so developers will see the SVG is malformed, and can optionally reject it) but eventually, one day, render it anyway.

However, I cannot understand the author's original intent. I understand that "move to 0" means "move to (0,0)", but ... WTF does "move to 9" mean?? Spec doesn't allow that at all AFAICS :(.

(Unless there's some additional info somewhere I'm missing)

PR #313, opened 2015-04-16
Fix for SVGs which have radius larger than w/h

chuckhendo:
I had an error using SVGKit with an SVG generated by Sketch. I noticed it was because the radius of one of the paths was the same as the width of the element. According to MDN, if the radius is greater than half of the width/height of the element, then is should be interpreted as half of the width/height of the element. This adds that functionality. 

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/rx


adamgit:
Don't use MDN - the real spec is here: 
http://www.w3.org/TR/SVG/shapes.html#RectElement

On quick glance it looks like MDN has copy pasted a small part (but not 
all!) of the official spec, so the effect is the same.

But for future reference ... the /TR/SVG/ site is best :)

Issue #214, opened 2014-09-22
SVGKLayer was deallocated while key value observers were still registered with it - iOS8.0


bsousa156:
Hi guys,

Thanks for the amazing work on SVGKit. 

There is a bug on iOS8.0 that makes the app crash. This also manifests with warnings on iOS7+.

When using SVGKLayeredImageView, once it gets deallocated it sends the following exception:

"**\* Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'An instance 0x7f847d8b8970 of class SVGKLayer was deallocated while key value observers were still registered with it. Current observation info: <NSKeyValueObservationInfo 0x7f847d8b92a0> (
<NSKeyValueObservance 0x7f847d884570: Observer: 0x7f847d8b8970, Key path: showBorder, Options: <New: YES, Old: NO, Prior: NO> Context: 0x0, Property: 0x7f847d894250>"

While on iOS7 the same happens but it does not break the app. I'm aware of a similar issue on the observers added to the DOMTree.viewport and i'm removing those on dealloc. But it does not solve the SVGKLayer issue.

adamgit:
Pull request #212 allegedly fixes this - but it does it by silently dropping the thrown exception. I can't merge that - it MUST at least log an Info or Warning log message.

This has also been discussed in #183, #184, and #187 - but each time someone brings this up, and we push for "OK, but we need a general solution, not a hack. Have you tried this in the other cases? Do you know this works for sure, or is it just a lucky guess that works on one single file?" etc, we never hear back.

Since no-one is willing to test this properly, I'd accept:
- combine #184 and #212 : use the #184 tactic of simply moving the line of code (!), and combine it with #212's guard (that protects against "this might still crash in rare cases"), but with a DDLogWarning for those cases ... so that if this DOES NOT fully fix the problem, we still at least get a warning.

...but I don't have iOS8 so I can't even test it.

bsousa156:
Hi Adam,

Thanks for the quick response. About iOS8, don't you have xcode6/ iOS8 simulator?

adamgit:
Testing on simulator isn't enough. Simulator uses a faked version of Quartz
/ CoreGraphics / CoreAnimation. There are MAJOR differences between the sim
and device here!

There are occasional issues (and this is exactly the kind of thing!) that
only trigger on device, or only on Simulator.

bsousa156:
[...]

PR #345, opened 2015-07-02
Fixes issue #214 for master-ARC branch (crashes on KVO when exiting a…

dovo329:
… viewcontroller that uses SVGKit)


adamgit:
I thought ARC ignored dealloc calls?

MaddTheSane:
`dealloc` is called when the retain count of the object reaches zero. In ARC code, many classes don't need the `dealloc` method as ARC automatically handles the memory clean-up a lot of `dealloc` methods are used for.

That said, there are a few classes that requires additional clean-up.

adamgit:
This pull request does NOT do what it claims to.

Instead, it adds support for the pixel densities for iPhone 6 - this is already in the main branch (2.x), and should be cross-ported / merged from there.

I'm closing this one, as there's been no activity for 6 months and it's incorrect in the first place. Re-open if you have changes.

PR #353, opened 2015-08-01
Create custom logging context for SVGKit

edave:
This PR sets up a [custom logging context](https://github.com/CocoaLumberjack/CocoaLumberjack/blob/master/Documentation/CustomContext.md) for the CocoaLumberjack logging. Doing this allows projects which use SVGKit and CocoaLumberjack to filter out SVGKit's logging by creating a [custom ContextFilter](https://github.com/CocoaLumberjack/CocoaLumberjack/tree/master/Demos/ContextFilter).

The only change for SVGKit is that the macros below should be used instead of DDLogger's built-in macros:

```
SVGKitLogError
SVGKitLogWarning
SVGKitLogInfo
SVGKitLogDebug
SVGKitLogVerbose
```

(all of the existing DDLog macros have been replaced with these SVGKit loggers)

All of these macros behave the same as their DDLogger counterpart, with the exception that a custom context (`SVGKIT_LOG_CONTEXT`) is set as part of the logging.

Developers using SVGKit in other projects can then filter out messages based on `SVGKIT_LOG_CONTEXT`. For example:

```
#import "SVGKit.h"
#import <CocoaLumberjack/CocoaLumberjack.h>

@interface CVGLoggingContextFilter : NSObject<DDLogFormatter>

@end

@implementation CVGLoggingContextFilter

- (NSString *)formatLogMessage:(DDLogMessage *)logMessage
{
    if (logMessage->_context == SVGKIT_LOG_CONTEXT)
    {
        // We can filter this message by simply returning nil
        return nil;
    }
    else
    {
        // We could format this message if we wanted to here.
        // But this example is just about filtering.
        return logMessage->_message;
    }
}

@end

```

On a side note, we tested these changes against the most recent version of CocoaLumberjack, so it may not work if you have an older version of that framework installed.

### DIFF
--- a/SVGKit-iOS.xcodeproj/project.pbxproj
+++ b/SVGKit-iOS.xcodeproj/project.pbxproj
@@ -265,13 +265,6 @@
 			remoteGlobalIDString = 2A37F97718E37270009FAAA7;
 			remoteInfo = CocoaLumberjack;
 		};
-		D320E8CA1ACCFA1F00BD4E72 /* PBXContainerItemProxy */ = {
-			isa = PBXContainerItemProxy;
-			containerPortal = D320E8C31ACCFA1F00BD4E72 /* Lumberjack.xcodeproj */;
-			proxyType = 2;
-			remoteGlobalIDString = 2A37F9DA18E37536009FAAA7;
-			remoteInfo = LibTest;
-		};
 		D320E8CD1ACCFAA700BD4E72 /* PBXContainerItemProxy */ = {
 			isa = PBXContainerItemProxy;
 			containerPortal = D320E8C31ACCFA1F00BD4E72 /* Lumberjack.xcodeproj */;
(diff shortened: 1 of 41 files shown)


seltzered:
this....this is awesome.

adamgit:
I'd like to push a new point-release of SVGkit very soon. Logging has improved quite a bit (not least: upgraded to Lumberjack 2.x is a pretty big change!). I need brief upgrade-path instructions.

As I understand it, this patch is complementary to the one(s) that allow you to build two versions of Lumberjack in same project? (using the compile-time flag in Lumberjack that they added specifically so that libraries like ours could embed LJ _as well as_ the applications that use our library separately building it)

edave:
I did not test it with building it multiple versions of Lumberjack, but it should just fine with that based on what I read about the custom context filtering. I believe that feature has been around for a long time because all their examples for custom context filtering were based on what looked to be the 1.X version of Lumberjack.

Issue #369, opened 2015-10-21
Add tvOS Cocoapods support


Jeehut:
I'd like to use this framework in a tvOS project using Cocoapods but in order to do this you need to explicitly state this project is made for tvOS, too.

For further details see [this blog post](http://blog.cocoapods.org/CocoaPods-0.39/).

adamgit:
If they're saying library authors "must" change their source to fit the needs of an external system, that's absurd. If you need to do this, can't you just edit your
CP file?

If not, it sounds to me like a good reason not to support CP.

seltzered:
I believe all that's needed is a change to the podspec file - @Dschee could you try adding in the modifications to the podfile and test it on tvOS? I recommend first creating your own fork, then in your podfile list reference svgkit to the local folder path. For example, I use svgkit (for an OSX app) via cocoapods, but the line in my podfile looks like this: 

```
pod 'SVGKit', :path => '../SVGKitMTSFork/'
```

As a reference to help in editing the file, check out the podspec on maddthesane's fork (see https://github.com/MaddTheSane/SVGKit/blob/master/SVGKit.podspec ) - there we do something similar in order to support OSX, and the podspec has helped in filtering out sources that are osx vs. ios specific. I'm expecting since tvOS / iOS are so similar, all you'll really need is the single deployment_target line. If all works definately send a pull req!

Jeehut:
@seltzered Thanks for the help, I'll give it a try and report the results.

Jeehut:
I tried and it didn't work in my case. So I'm opting for a manual integration via git submodules now. If anybody wants to try on their own project, here's how to add my fork:

– make sure you have the latest version of cocoapods installed (`gem install cocoapods`)
– simply add this to your Podfile: `pod 'SVGKit', git: 'https://github.com/Dschee/SVGKit'`

If it turns out other people seem to have no problem with it I'll be happy to send a pull request.

dreampiggy:
Fix now. You can try using the CocoaPods version 2.1.0 via https://cocoapods.org/pods/SVGKit

Issue #283, opened 2015-03-16
Can you please add XCode6 framework build target for this project?


GasimGasimzada:
I am also trying to build this project as a Framework, which makes the integration so much cleaner. Currently, I am getting a weird error "Use of undeclared identifier 'ddLogLevel'"; so I am investigating. Will update this post, once I get this framework into a working condition. I would have wanted to make a pull request but there are information that I won't be able to commit (like Bundle Identifier etc).

adamgit:
Do you need a new bundle identifier for making this a framework?

GasimGasimzada:
AFAIK you do. I have already successfully compiled it. Importing the framework in the following way:

```
 import SVGKit 
```

Just makes it so much cleaner, with the framework being its own project. I will test it tonight and see if this works. And if it does, I will check to see if I can remove the bundle identifier from the project.

GasimGasimzada:
I got it to work finally. Here are my options:

```
I got it to work finally. Here are my options:
Target Name: SVGKitFramework
Product Name: SVGKit // this one is important!
Other Linker Flags: -ObjC

// to match the the project settings:
Info.plist File: XCodeProjectData/SVGKitFramework/Info.plist
Prefix Header: XCodeProjectData/SVGKitFramework/SVGKitFramework.pch

Header Search Paths: added /usr/include/libxml2
Objective-C Automatic Reference Counting: NO

Compile Files: All 100 items
Link Binary with Libraries:
    QuartzCore.framework
    Foundation.framework
    CoreGraphics.framework
    libxml2.dylib
Header Files: All 128 items
```

And usage is very simple:
1. First add SVGKit Project as a sub project
2. Add these settings in Build Phases:
   
   Target Dependencies: Add SVGKitFramework
   Copy Files: Add SVGKit.framework

PR #370, opened 2015-10-23
Support for packaging as iOS frameworks

Jeehut:
In order to get this project working in a tvOS project that I'm working on right now I needed to update/change a few things:
- [x] package as Framework (for easier inclusion in iOS & tvOS) – #283
- [x] added Cocoapods tvOS support – #369
- [x] added setBundle: method for imageNamed: method (to make it work with frameworks including assets)

I'm currently testing the changes in the project and haven't started tvOS integration yet, so there might still be some glitches. But since it's basically working for me now I felt like making the pull request already. I'll keep you updated on the progress.

_UPDATE:_
After input from @adamgit this pull request is not intended to get tvOS working completely. This is for making framework packaging possible only (with some minor tvOS related changes included already).


Jeehut:
This pull request is supposed to fix https://github.com/SVGKit/SVGKit/issues/369 and https://github.com/SVGKit/SVGKit/issues/283.

adamgit:
We tried repackaging main project as a framework a while back, and we 
failed. Couldn't get it to work.

If you managed to get that working, is it possible to submit as a separate 
pull request? Because that's a huge change (big improvement, IMHO!) all on 
its own.

Then submit the tvos/podspec changes as an immediate followup.

there are more people who can test the framework bit than can test the tvos 
stuff, so.I'd like to get someone reviewing it first asap.

Jeehut:
@adamgit The tvOS bit doesn't prevent anybody from testing the iOS framework bit and it didn't include any changes except marking the podspec as tvOS compatible (what won't do any harm) and creating a tvOS target alongside an iOS target. Therefore I don't think it's necessary to keep these things apart from each other.

I agree though, that we should make sure the framework integration works first. So, I have just updated the name of this pull request accordingly and removed the tvOS-related TODOs. I think we can keep the tvOS target and the one line within the podspec, don't you agree? I haven't done any other tvOS-related changes yet and will send those with a different pull request then if needed.

If you think we should remove the minor tvOS-related changes, let me know.

Jeehut:
Btw: I say things are working for me, but what I mean is: Everything is building. I have done more changes on my project than only including this project as a framework and it's not all working yet. But I can't tell if it's an issue with this framework or the project, so others will need to test the framework integration in their projects, too.

adamgit:
Yep, exactly, I'd like to get a few different heavy users testing this with 
their different set ups

Source/Sources/SVGKSourceLocalFile.m
```diff
@@ -52,7 +60,14 @@ +(SVGKSourceLocalFile*) internalSourceAnywhereInBundleUsingName:(NSString*) name
 	
 	/** First, try to find it in the project BUNDLE (this was HARD CODED at compile time; can never be changed!) */
 	NSString *pathToFileInBundle = nil;
-	NSBundle *bundle = [NSBundle mainBundle];
+
```
adamgit, line 41: make this  a parameter to the method, rather than adding global state

Jeehut, line 41: Just did in 36314710829fce676c2d1b7464d6dec32ec76e27.

adamgit:
@ ... everyone:

Please try this out, test it with your projects, and report back ASAP if any problems / crashes / failures. If it breaks bulding for anyone, I'll revert.

adamgit:
@Dschee - what are the install / usage steps for people wanting to use this directly as framework?

I'm guessing:
1. Open SVGKit-iOS project (the original one)
2. Build
3. ... there's a Framework output somewhere?

Problem is ... I'm not seeing it

Jeehut:
@adamgit: I'm not using this framework anymore but what I basically did back then when I was using SVGKit was to just drag and drop the `SVGKit-iOS.xcodeproj` to my project navigator, then I could choose `SVGKit.framework` within `Linked Frameworks and Libraries` in my targets `General` tab. Does this help?

PR #373, opened 2015-10-29
Potential overrelease in setter methods

benpious:
There are several setter methods which  use the following pattern:

```
- (void)setObject:(id)o {
  [_oldObject release];
  _oldObject = o;
 [o retain];
 }
```

if the setter were called twice with the same argument for `o`, then, unless `o` had been retained by some other object in the interim, `o` would have a reference count of zero after the second `release`, and be deallocated. Having little experience with this library, I'm not certain if this would ever happen in practice, but even if it isn't an issue now, it might be prudent to fix it now, in case something changes in the future.

### DIFF
--- a/Source/DOM classes/Core DOM/CSSPrimitiveValue.m
+++ b/Source/DOM classes/Core DOM/CSSPrimitiveValue.m
@@ -224,9 +224,10 @@
 
 -(void)setCssText:(NSString *)newCssText
 {
-	[_cssText release];
+  NSString *oldCSSText = _cssText;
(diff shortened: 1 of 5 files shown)


adamgit:
I'm getting merge conflicts here :(.

Normally, I think the pattern is to do:
1. "if( newValue == oldValue ) return"
2. ... rest of method

adamgit:
It's been 3 years and no followup - I couldn't merge back then because of the conflicts - so I'm closing this now. It seems to have been OK without the patch.

PR #376, opened 2015-11-04
Supporting iPhone 6 & 6S, 6 Plus & 6S Plus

anuragt84:
Adding the pixel density for iPhone 6, 6S and 6 Plus, 6S Plus

### DIFF
--- a/Source/DOM classes/SVG-DOM/SVGLength.m
+++ b/Source/DOM classes/SVG-DOM/SVGLength.m
@@ -143,9 +143,15 @@ static float cachedDevicePixelsPerInch;
 	
 	if( [platform hasPrefix:@"iPhone4"]
 	|| [platform hasPrefix:@"iPhone5"]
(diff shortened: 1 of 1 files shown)


anuragt84:
The spacing seems off in the diff, but it is fine in XCode.

MaddTheSane:
The different spacing is because (it looks like) the old code uses tabs, whereas your new code uses spaces. Xcode defaults to tabs being equivalent to 4 spaces; on GitHub, it's similar to 8.

anuragt84:
Can you tag the 1.x branch with 1.2.1?

adamgit:
1.x branch has been retired a long time ago. We shouldn't be doing commits 
to it - if this was a 1.x commit, apologies, I wouldn't have merged if I'd 
noticed.

Really, if you're on 1.x, get off it as fast as you can. 2.x is much much 
better!

PR #390, opened 2015-12-03
Fix Issue #214

sciutand:
Fixes issues #214 for SVGKFastImageView by tracking observers being added and make sure they are only added once and removed if added during dealloc.
### Note

This only fixes SVGKFastImageView it does not touch other classes that might have similar issues.

### DIFF
--- a/Source/UIKit additions/SVGKFastImageView.m
+++ b/Source/UIKit additions/SVGKFastImageView.m
@@ -152,7 +152,7 @@
   if (!self.didRegisterInternalRedrawObservers) return;
 	[self removeObserver:self  forKeyPath:@"layer" context:internalContextPointerBecauseApplesDemandsIt];
 	[self.layer removeObserver:self forKeyPath:@"transform" context:internalContextPointerBecauseApplesDemandsIt];
-	//[self.image removeObserver:self forKeyPath:@"size" context:internalContextPointerBecauseApplesDemandsIt];
+  self.didRegisterInternalRedrawObservers = false;
 }
 
 -(void)setDisableAutoRedrawAtHighestResolution:(BOOL)newValue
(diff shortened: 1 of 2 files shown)


Source/UIKit additions/SVGKFastImageView.m
```diff
@@ -9,6 +9,8 @@
 @interface SVGKFastImageView ()
 @property(nonatomic,readwrite) NSTimeInterval timeIntervalForLastReRenderOfSVGFromMemory;
 @property (nonatomic, retain) NSDate* startRenderTime, * endRenderTime; /*< for debugging, lets you know how long it took to add/generate the CALayer (may have been cached! Only SVGKImage knows true times) */
+@property (nonatomic) BOOL didResigesterObservers, didResigesterInternalRedrawObservers;
```
adamgit: Please fix the typo: "didResigesterObservers" should be: "didRegesterObservers"

"didResigesterInternalRedrawObservers" should be "didRegesterInternalRedrawObservers"

Source/UIKit additions/SVGKFastImageView.m
```diff
@@ -130,20 +130,26 @@ - (void)setImage:(SVGKImage *)image {
...
+  if (self.didResigesterInternalRedrawObservers) return;
+  self.didResigesterInternalRedrawObservers = true;
 	[self addObserver:self forKeyPath:@"layer" options:NSKeyValueObservingOptionNew context:internalContextPointerBecauseApplesDemandsIt];
 	[self.layer addObserver:self forKeyPath:@"transform" options:NSKeyValueObservingOptionNew context:internalContextPointerBecauseApplesDemandsIt];
-	//[self.image addObserver:self forKeyPath:@"size" options:NSKeyValueObservingOptionNew context:internalContextPointerBecauseApplesDemandsIt];
 }
 
 -(void) removeInternalRedrawOnResizeObservers
 {
+  if (!self.didResigesterInternalRedrawObservers) return;
```
adamgit: This method should ALSO set .didRegisterInternalRedrawObservers back to FALSE at the end, no?

adamgit:
After all the workarounds we've tried, I believe this is the simplest possible solution - as you've done, we need to track status of registration, Apple's API's and the confusing set of edge-cases are too messy on their own.

c.f. comments

Source/UIKit additions/SVGKFastImageView.m
sciutand: Yes sorry not sure what I was thinking.

Source/UIKit additions/SVGKFastImageView.m
sciutand: Indeed, good catch

sciutand:
@adamgit fixed PR based on feedback.

PR #392, opened 2015-12-07
Attempting to use Xcode6 frameworks

janeabraham:


adamgit:
Please don't attempt to merge this - it won't work.

The commit you're asking to merge _never worked_ - I tried to make it work, but it needs contributions from other people, because I keep running into bugs in Apple's code :(.

PR #406, opened 2016-01-11
Demo project build fixes

alexandre-g:
Renamed 3rd party frameworks folder to 3rd-party-frameworks as Xcode doesn't put required quotes around framework search paths on drag and drop import

The `#if V_1_COMPATIBILITY_COMPILE_CALAYEREXPORTER_CLASS` conditionals in DetailVC are somewhat messy, but I am not sure which path @adamgit would like to take this? Maybe take this code out of the demo altogether if deprecated?

PS. not too experienced with github - hope this doesn't break more than it fixes )


adamgit:
Yes, I'd like to remove the CALayerExporter entirely, just didn't want to delete it without warning. Maybe wait a few weeks then kill it.

Obviously it's all in 2.0.0 release, so if anyone desperately needs it, they can easily add it themselves.

PR #407, opened 2016-01-12
zoom fixes

alexandre-g:
fixes zoom for `SVGKFaskImageView` (delegate wasn't set on scroll view)
and proposes a solution to zoom `SVGKLayeredImageView`

### DIFF
--- a/Source/SVGKImage.h
+++ b/Source/SVGKImage.h
@@ -290,9 +290,6 @@ typedef void (^SVGKImageAsynchronousLoadingDelegate)(SVGKImage* loadedImage, SVG
 /*! returns all the individual CALayer's in the full layer tree, indexed by the SVG identifier of the SVG node that created that layer */
 - (NSDictionary*) dictionaryOfLayers;
 
-/*! returns all the individual CALayer's in the full layer tree as an array */
-- (NSArray*) arrayOfLayers;
-
 #pragma mark - Useful bonus methods, will probably move to a different class at some point
 
 /** alters the SVG image's size directly (by changing the viewport) so that it will fit inside the specifed area without stretching or deforming */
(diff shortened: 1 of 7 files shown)


Source/SVGKImage.h
```diff
@@ -290,6 +290,9 @@ typedef void (^SVGKImageAsynchronousLoadingDelegate)(SVGKImage* loadedImage, SVG
 /*! returns all the individual CALayer's in the full layer tree, indexed by the SVG identifier of the SVG node that created that layer */
 - (NSDictionary*) dictionaryOfLayers;
 
+/*! returns all the individual CALayer's in the full layer tree as an array */
+- (NSArray*) arrayOfLayers;
```
adamgit: Why is this needed?

Doesn't dictionaryOfLayers provide 100% of this already?

I'm OK with adding features, but I don't want to add anything to SVGKImage.h unless it's essential -- I really want us to move features OUT of SVGKIimage.h and into their own classes!

alexandre-g: Thought you might not like that :) To be honest, the reason I've added it is because dictionaryOfLayers does not work at the moment, and I wasn't able to fix it right off.
Will submit a new pull request when I do.

adamgit: If you can remove that from this PR, I'll merge it - the rest all looks
fine.

If you have a specific problem with dictOfL, point me to an issue / make an
issue and I'll look at it - I use it quite often, and last time I tried it
was working fine. If someone's accidentally broken it, I might be ablt to
spot the problem quickly.

adamgit:
I can't merge with the added Array. I'd suggest removing that sooner rather than later, or else there's a high probability someone else will merge a change that makes this PR stop working, and then it'll get discarded :(.

alexandre-g:
Hey, my OSX got corrupt beginning of the week and recovery didn't help. Just got time to try and bring it back to life. Will look into the dictionaryOfLayers issue and resubmit new PR once im back in the game.

PR #417, opened 2016-01-27
Fix a compiler warning: Variable 'valueAsInches' may be uninitialized when used here

cxa:


adamgit:
Set it in the "default:" case instead - same effect on compiler, but is more  correct and easier to debug.

PR #419, opened 2016-02-04
Fix exception throw for empty G Element

badjeff:


adamgit:
If you set a frame to CGRectZero that specifcally means "a 1 pixel frame at (0,0)" in Apple's library.

(years ago, the main code originally did what you've coded, and it broke many SVGs. Most cases we managed to think of different ways of handling the situiation, but this one we never had a workaround for)

That breaks rendering badly: it is moving and resizing every layer (and because these are nested: it is potentially moving ALL layers and resizing ALL layers in the SVG :( ).

Also: never remove an Assertion unless you are 100% sure you have fixed the problem for ALL situations - which this doesn't? Even then you should probably convert it to a Log message until you know for sure that you didn't miss any edge cases.

Finally: we're using Cocoa Lumberjack, so your NSLog needs to be the Lumberjack 2.0 equivalent instead of NSLog.

adamgit:
NB: unless Apple has made an OS-breaking change to their library recently and CGRectZero has been redefined, in which case I'd like to see a test-case that demonstrates this. They'd be breaking potentially every iOS and OS X app published over the last 5 years. Apple could easily do that (they've done it before), but it's not something they do often.

badjeff:
In my case, Apple does not break my published executable. It breaks all re-compiled app in Xcode 7.2 + Swift 2.0 on iOS 9 (works on iOS 8). Anyway, solved this by method swizzling, if exception is caught.
NB: shall be happy to have a preprocessor definition for switch between Assertion and Warning.

PR #431, opened 2016-03-16
Updates so Carthage will now work

equinox2k:
Small update so Carthage now works pointing to repo.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,0 @@
-[submodule "calayer-exporter"]
-	path = calayer-exporter
-	url = git://github.com/reklis/calayer-exporter.git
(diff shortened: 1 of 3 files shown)


equinox2k:
A small request in order for carthage to pick up latest version, you need to tag regularly with a version e.g. after key fixes + features. Is it possible to tag the latest commit for example as 2.0.1 or what ever you feel best as currently can only point to 2.0 since that is the latest tag.

Many thanks

adamgit:
I suggest submitting fixes to Carthage so that it can support real-world development, instead of trying to force projects into changing their process to fit Carthage.

This is a small project, and it's not complete yet - there's a lot of spec (animations!) missing. Until then, releases are arbitrary, and any time spent making artificial "releases" to satisfy a build-tool is time that fixes and features are not being added - and that time will never be won back :(.

adamgit:
(NB: 3 different maintainers - including myself - have tried making releases and tags, and it proved too much effort / low priority / wasted time for too little benefit)

If you volunteer to run the tagging, that would be great. But having seen two people drop out after I got them setup, I'd want to be sure it was going ahead before burning time on it again.

markst:
Not sure if it's a carthage issue or SVGKit, but this is the error I get:

```
*** Cloning SVGKit
*** Checking out SVGKit at "2.0.0"
Parse error: expected submodule commit SHA in output of task (ls-tree -z 2.0.0 calayer-exporter) but encountered: 
```

markst:
@equinox2k is it mandatory that repo's tag their releases for carthage to work?

PR #444, opened 2016-05-12
Fixes #434: Acknowledge style attributes on root <svg> tag

### LINKED ISSUE #434: 2.x: Fill attribute on root <svg> element ignored
I've been using Google's Material Design icons: https://design.google.com/icons/

Google establishes a fill attribute on the root <svg> element, and only lists a `fill="none"` on paths without a fill.

Using SVGKit, it seems to expect a fill attribute for each path and doesn't pay attention to a fill attribute on the root element.

cpboyd:
Fixes issue #434.
`<SVG>` tags may still contain stylable properties like fill.
Ref.: https://www.w3.org/TR/SVG11/struct.html#SVGElement

### DIFF
--- a/Source/DOM classes/Unported or Partial DOM/SVGElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGElement.m
@@ -567,9 +567,9 @@
 	 
 	 http://www.w3.org/TR/SVG/propidx.html
 	 
-	 For each of those, the implementaiton is the same.
+	 For each of those, the implementation is the same.
 	 
-	 ********* WAWRNING: THE CURRENT IMPLEMENTATION BELOW IS VEYR MUCH INCOMPLETE, BUT IT WORKS FOR VERY SIMPLE SVG'S ************
+	 ********* WAWRNING: THE CURRENT IMPLEMENTATION BELOW IS VERY MUCH INCOMPLETE, BUT IT WORKS FOR VERY SIMPLE SVG'S ************
 	 */
     NSString* localStyleValue = [self.style getPropertyValue:stylableProperty];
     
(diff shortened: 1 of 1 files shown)


adamgit:
Oh, I see the problem now.

This won't work, needs more thought: you can have _multiple_ SVG nodes in a single doc. "SVG" does _not_ mean "root" - it means local root.

The rules are that you MUST NOT go beyond a local root.

So ... I guess we need to ask "is current node SVG? if so ... do not continue" ?

cpboyd:
@adamgit It still shouldn't go past the `<SVG>` or `<G>` tags.

I left the while loop intact, to break when either of those tags are reached.

The only code change was to prevent pre-emptively returning `nil` on an `<SVG>` tag.
From ([line 636](https://github.com/SVGKit/SVGKit/blob/2.x/Source/DOM%20classes/Unported%20or%20Partial%20DOM/SVGElement.m#L636)):

``` objc
        if( parentElement == nil
           || [parentElement isKindOfClass:[SVGSVGElement class]] )
        {
            return nil; // give up!
        }
```

to

``` objc
        if( parentElement == nil )
        {
            return nil; // give up!
        }
```

cpboyd:
For reference, the while loop I mentioned (on [line 628](https://github.com/SVGKit/SVGKit/blob/2.x/Source/DOM%20classes/Unported%20or%20Partial%20DOM/SVGElement.m#L628)):

``` objc
        Node* parentElement = self.parentNode;
        while( parentElement != nil
              && ! [parentElement isKindOfClass:[SVGGElement class]]
              && ! [parentElement isKindOfClass:[SVGSVGElement class]])
        {
            parentElement = parentElement.parentNode;
        }
```

This results in the code looping back to the element's containing `<G>` or `<SVG>` tag, but the `if` statement changed in this commit would always return `nil` if an `<SVG>` tag was reached.

PR #465, opened 2016-08-13
Fix for noticable performance hit due to fonts loading

RenGate:
When `UIViewControllers` that contain `SVGKLayeredImageView` are instantiated from Storyboards, a no image placeholder is automatically set during instantiation and a font used to render "Missing SVG" is loaded using `CTFontCreateWithName`.
Because `UIViewControllers` are usually instantiated during transition, such font loading causes noticeable lag and makes user experience worse. During testing using Time Profiler on iPhone 5s (iOS 9.3.3) I found out that such font loading increased the view controller loading time by 2 seconds. After removal of text layers from placeholder image, instantiation time dropped to usual 0.4 s.


adamgit:
Suggestions for something we can put there instead?

RenGate:
Other option is to outline text in graphics editor instead of removing it completely. The loading overhead will be bigger than in case of a simple rectangle, but still far less than rendering text using CoreText.

adamgit:
Sounds like a good idea.

And maybe ... we add a #define that lets you comment it out if you really
don't want it?

But I feel the default needs to be "obvious" for new users who make a
misake and dont' realise what's happened.

RenGate:
Sounds good.
I've added the outlined text and placed the macro check, so now we can disable placeholders on `SVGKLayeredImageView` by adding preprocessor macro in build settings.

lastMove:
👍

PR #473, opened 2016-09-18
Update OTHER_SDK_TO_BUILD regex for iOS 10/Xcode 8

rawrmaan:
Building the static library against iOS 10 base SDK current fails due to this naive regex in the Run Script build phase written by someone who, I guess, thought iOS would never get past version 9.0:

`SDK_VERSION=$(echo ${SDK_NAME} | grep -o '.\{3\}$')`

This results in `SDK_VERSION` being `0.0` instead of `10.0` when building against iOS 10 Base SDK, causing a build error along the lines of `xcodebuild: error: SDK "iphonesimulator0.0" cannot be located`

My fix is just as naive as the initial regex--we just check 4 characters in the SDK_NAME string instead of 3:

`SDK_VERSION=$(echo ${SDK_NAME} | grep -o '.\{4\}$')`

Maybe someone with better Regex skills than I can make this work for 3-char and 4-char SDK_NAME strings so this PR doesn't break iOS <10 builds


adamgit:
Nice spot!

FYI That script was originally a quick hack from someone on SO.com, and I merely added to it based on trial and error with xcode's undocumented build process, and kept adding more erorr handling and stuff until it was robust.

Technically ... I think Apple could legitiamtely break it tomorrow - they've never committed to this naming scheme, nor documented it (IIRC).

rawrmaan:
@adamgit Just wanna make sure I'm completely clear that the build will now break if you build against iOS <10. I believe the correct regex to work for iOS `x.x` and `xx.x` would be

`SDK_VERSION=$(echo ${SDK_NAME} | grep -o '.\{3,4\}$')`

But I haven't tested it.

adamgit:
Nope, it's picking N characters from the end of the line. You cannot say
"pick 3 or 4 characters from end of line", it will take the largest number
you ask for.

The whole regexp needs to be changed into something that only matches
digits.

adamgit:
This should work, tested on a few examples from command line

grep -o '\d{1,2}.\d{2}$'

PR #492, opened 2016-11-08
Fixed incorrect return type

synchromation:
Both imageWithContentsOfFileAsynchronously and imageWithDataAsynchronously return SVGKParser not SVGKImage


adamgit:
The interfaces are correct. They MUST return SVGKImage

synchromation:
I'm probably missing something (just trying to fix some warnings flagged by Xcode 8.1), but imageWithContentsOfFileAsynchronously returns the image in the completion block, but the actual method returns the value from imageWithSource which returns a SVGKParser I believe

adamgit:
Yep, so probably the method name needs to change.

eg parserParsedImageWith...

...but I don't know what the naming convention is for methods which use a 
block as result.

PR #502, opened 2016-12-12
Fixed wrong return type

werner77:


adamgit:
Someone made the same change recently, and I requested that they either change the method name too (since it CANNOT have this return type with that name!) and/or change the implementation to return void.

So, either:

SVGKParser* parserWithContentsOfFileAsynchronously:

or

void asynchronouslyParseImageWithContentsOfFile:

...the former would have least impact on existing code, and so I think is better? (but I'd appreciate input from others on this - either way it will definitely impact existing projects)

werner77:
An option could be to deprecate the current method, by leaving the API as is but always return nil.

Then add the "SVGKParser* parserWithContentsOfFileAsynchronously:" method that replaces the former (as put in the deprecation message).

Regards, Werner

PR #507, opened 2016-12-22
Changes in the Debug build configuration to improve building static library experience

vitalii-tym:
"Build Active Architecture Only" set to NO to make the resulting static lib include both architectures - for Simulator and Device.
"-fembed-bitcode" enables apps built with SVGKit to be submitted to AppStore without deselecting the "include bitcode" checkbox. Having bitcode included is better, because Apple will be able to optimise app footprint.

### DIFF
--- a/SVGKit-iOS.xcodeproj/project.pbxproj
+++ b/SVGKit-iOS.xcodeproj/project.pbxproj
@@ -2756,6 +2756,8 @@
 					"$(inherited)",
 					"\"$(SRCROOT)/Source/Core\"",
 				);
+				ONLY_ACTIVE_ARCH = NO;
+				OTHER_CFLAGS = "-fembed-bitcode";
 				PRODUCT_NAME = "$(TARGET_NAME).${DYLIB_CURRENT_VERSION}";
 				SKIP_INSTALL = YES;
 			};


adamgit:
The embed-bitcode I understand - but why change the Active architecture, does that even work?

Note the long shellscript that forces the building of the different architectures, then automatically creates the dual-binary (that Apple deleted from Xcode 5). Does Xcode 7 correctly create a dual-binary now? (last time I checked, it seemed not to).

vitalii-tym:
I don't know why, but script didn't do dual-binary for me. I always had either Simulator or Device depending on my selection in "Device". So I had to make this setting to have both binaries included.
It is important to note that:
1) I'm still on XCode 7.3.1. Couldn't check how it works on XCode 8, probably not an issue any more.
2) Last time I built SVGKit was October 26, which was before this commit: https://github.com/SVGKit/SVGKit/commit/e01c401035b9731b2500345f83b4bda0b76b9146
(so I'm not sure whether the updates in script in that commit could fix the dual-binary issue I had)
Probably this pull request can be set on hold until someone ensures it doesn't break things on XCode 8. I will be upgrading somewhere only in 1-2 weeks (will check myself if that happens first).

adamgit:
That commit is 100% required for building against iOS 10.x or above (without it, script only builds iOS 2-9 :))

If it works on xcode 7, I'm happy to merge - most teams I know are still on 7, or supporting 7 and 8, haven't switched to 8 completely yet.

Please pull latest, check whether the Octoer fix makes dual-binary work for you. If not, please raise an issue for it - it SHOULD work, and I'd like to fix it ASAP if not. But I don't have acces to a machine with xcode on it this week :(. So I can't check myself.

vitalii-tym:
... just had a quick look at the script and realized that creating both binaries was its main purpose.
So, probably some part of the script was failing for me because I remember it creating the "Debug-universal" folder correctly and having the lib file in there, but the file had only one architecture (it was ~10MB in size).

Rechecked the building process with that latest commit - everything went fine. Ends up in "Debug-universal" folder with the lib file ~20 MB in size as expected. With this file my project builds correctly for device and for simulator (It is only CocoaLumberjack that I have to switch back and forth as usual). 
I didn't check AppStore submittion though.

adamgit:
OK, so can you flip active arch back to what it was, and then I can merge this PR?

vitalii-tym:
Are you sure "Active Architecture Only" should be flipped back? 
It would result in the app producing not quite full dual-binary. For example when I build the static library for "Generic iOS Device", my project won't build for Simulator giving me this error:
```
ld: warning: ignoring file /Users/vitaliy/Documents/Projects/EasyJira/Traffic/Traffic-iOS/libSVGKit-iOS.2.0.0.a, missing required architecture x86_64 in file /Users/vitaliy/Documents/Projects/EasyJira/Traffic/Traffic-iOS/libSVGKit-iOS.2.0.0.a (3 slices)
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_SVGKImage", referenced from:
      objc-class-ref in ProjectsViewController.o
      objc-class-ref in TasksViewController.o
      objc-class-ref in IssueDetailsViewController.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```
This was on Xcode 7 but I've also tested that Xcode 8 works the same.

vitalii-tym:
I have verified on Xcode 8 that only having "Build Active Architecture Only" set to NO produces correct dual-binary.

vitalii-tym:
... I also verified that my app with SVGkit was corectly uploaded to AppStore with symbols and bitcode included.

Issue #500, opened 2016-12-11
Upgrade to swift 3 fails because of CocoaLumberjack 2.x dependency


kishykumar:
Hi, 

I am trying to move my project to swift 3 and getting the following error when I try to update CocoaLumberjack to swift3 branch. 

=============
$ pod install
Analyzing dependencies
Pre-downloading: CocoaLumberjack from **https://github.com/CocoaLumberjack/CocoaLumberjack, branch swift_3
Pre-downloading: SVGKit from https://github.com/SVGKit/SVGKit.git, branch 2.x**
[!] Unable to satisfy the following requirements:

**- CocoaLumberjack (~> 2.x) required by SVGKit (2.0.1)**

None of your spec sources contain a spec satisfying the dependency: CocoaLumberjack (~> 2.x).
============

Is there a swift3 version of SVGKit? OR Can we please remove this dependency somehow?

adamgit:
That suggests the podspec file is wrong, no?

Anyway, lumberjack is built in to the project, so it shouldn't be listed in 
podspec. I guess the podspec is out of date, maybe just change it by hand.

PR #511, opened 2017-02-15
Change CocoaLumberjack to 3.x

### LINKED ISSUE #481: Use of undeclared identifier LOG_LEVEL_DEF
Using latest Xcode I'm getting this error.  Can bypass this if I `#define` them to `0`, but doesn't seem like a long term plan

cpboyd:
Fixes #481 and #500

### DIFF
--- a/SVGKit.podspec
+++ b/SVGKit.podspec
@@ -20,7 +20,7 @@ Pod::Spec.new do |s|
   s.tvos.deployment_target = '9.0'
   s.libraries = 'xml2'
   s.framework = 'QuartzCore', 'CoreText'
-  s.dependency 'CocoaLumberjack', '~> 2.x'
+  s.dependency 'CocoaLumberjack', '~> 3.x'
   s.prefix_header_file = 'SVGKitLibrary/SVGKit-iOS/SVGKit-iOS-Prefix.pch'
   s.requires_arc = true
   s.xcconfig = {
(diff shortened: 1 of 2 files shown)


cpboyd:
This updates the podspec's CocoaLumberjack to 3.x and corrects the import to work with the Swift Framework (which resulted in #481 starting in CocoaLumberjack 2.4).

cpboyd:
@Streebor 
I don't know if there's a good workaround until https://github.com/CocoaLumberjack/CocoaLumberjack/issues/778 is fixed.

CocoaLumberjack hasn't supported the non-framework `#import <CocoaLumberjack/CocoaLumberjack.h>` since 2.3.

Is there a specific reason you're against using frameworks?

There haven't yet been any commits related to any other issue, so you could possibly specify commit 6a89432 in your podfile for now.
e.g. `pod 'SVGKit', :git => 'https://github.com/SVGKit/SVGKit.git', :commit => '6a89432'`

Streebor:
@cpboyd Thank you for your quick response!

There are other older libraries what we depend on that do not support frameworks. But even bigger problem is our app's extension for Messages. App extensions cannot have frameworks  because the whole app will be rejected.

cpboyd:
@Streebor 
Regarding app extensions (and assuming you're using Cocoapods):
https://github.com/CocoaPods/CocoaPods/issues/4203 and a related merge  https://github.com/CocoaPods/CocoaPods/pull/5732

I don't know if those would apply to a Messages app extension without a host app, though.

There's also https://github.com/CocoaPods/CocoaPods/issues/5594 and  https://github.com/CocoaPods/CocoaPods/issues/5860 which do specify Messages extensions and embedding frameworks.

I'm not really sure I understand what you mean by older libraries not supporting:
Do you mean that they don't work with the newer CocoaLumberjack?

I mean, there's some people like #500 that want/need the newer version.

PR #519, opened 2017-03-28
Elliptic parser fix

aronkool:


adamgit:
I don't have a unit test for this, but the code looks OK to me - it only affects the elliptical-arc primirtive, although I remember there are some tricky edge-cases for when whitespace and commas can appear (because of the SVG rules that allow authors to compress by removing inferable data).

Is there a file in the project that shows this working / not working? (should be one of the W3C spec files, I imagine?)

aronkool:
Thank you for the quick response and check.
I do agree that there should be a Unit Test for this, but unfortunately I didn't had the time to build one. The file that caused the error and the one I used for debugging and testing, is the following:
https://talk.parro.com/assets/img/Parro-App-Agenda-Images_5.svg

PR #520, opened 2017-04-03
Correcting return types.

avinashdongarwar:
This fixes the last 2 warnings in following screenshot.
screen shot 2017-04-03 at 9 31 40 am


adamgit:
I've commented on this a few times.

It is not a fix to merely rename the return-type. Either the method must be updated to support the claimed return-type, or the return-type must be changed in the interface AND the method renamed.

(you can't have a method "image..." that returns something that's not an image).

I'd accept either the above approaches, but not a simple renaming of the return-type.

PR #521, opened 2017-04-03
Return image returned via asynchrounsly parser function.

avinashdongarwar:
An alternative approach to my earlier PR https://github.com/SVGKit/SVGKit/pull/520
This is based on following suggestion by @adamgit .
screen shot 2017-04-03 at 2 54 09 pm


adamgit:
This is a great idea to improve the old (broken) code.

I think we should keep both patched versions.

Can you re-do this as an EXTRA set of methods? So we end up with:

imageParser...
and
image...

...and developers can use either? Most cases will use the imageParser one (it's needed by most of the internal SVGKit methods), but for convenience, anyone can use the image... one.

avinashdongarwar:
@adamgit Modified the code as per your comments. Please re-review.

adamgit:
Thanks - great work!

PR #544, opened 2017-09-06
Remove obsoloete variables in SVGLength

vladdeSV:
`SVGLength` has some unused variables which made the impression that, in the debugger, sizes weren't loaded properly.

This fix removes those variables.

### DIFF
--- a/Source/DOM classes/SVG-DOM/SVGLength.h
+++ b/Source/DOM classes/SVG-DOM/SVGLength.h
@@ -45,10 +45,8 @@ typedef enum SVG_LENGTH_TYPE
 
 @interface SVGLength : NSObject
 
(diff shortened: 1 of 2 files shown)


vladdeSV:
I really suggest someone with more knowledge of this project tests and verifies that this fix doesn't introduce any new bugs.

adamgit:
I believe these are part of the SVG official spec, and that's why they're there.

Implementing / using them is then a TODO :), but they shouldn't be deleted.

adamgit:
As I suspected: if you click the link at the top of the header-file, it explains what these variables are. They are part of the spec. It's a bug that they're not being set, but it's not an accident that we have them!

PR #554, opened 2018-01-24
Fix NSData crash when url or path cannot be opened

teffi:
Added try-catch-finally to avoid fatal exception when loading image contents of a path or url. 

> Error internally in Apple's NSData trying to read from URL 'path to svg file'. Error = Error Domain=NSCocoaErrorDomain Code=256 "The file “svg filename” couldn’t be opened." UserInfo={NSURL=path to svg file}

Use:
`+[SVGKImage imageWithContentsOfURL:completion:]`
`+[SVGKImage imageWithContentsOfFile:completion:]`

### DIFF
--- a/SVGKitLibrary/SVGKit-iOS/SVGKit-iOS-Prefix.pch
+++ b/SVGKitLibrary/SVGKit-iOS/SVGKit-iOS-Prefix.pch
@@ -4,7 +4,7 @@
 
 #ifdef __OBJC__
 #import <Foundation/Foundation.h>
-@import CocoaLumberjack;
+#import <CocoaLumberjack/CocoaLumberjack.h>
 
 #define SVGKIT_LOG_CONTEXT 556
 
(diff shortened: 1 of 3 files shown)


adamgit:
Why have you created a new completion block for this?

The old block should be sufficient - but if not, please change ALL the instances, don't create a new, duplicate, typedef. Your new block forces peolpe to duplicate their code and handle both outcomes - and it also appears to discard critical information - you've lost ALL the parse data, no?

teffi:
Sorry. I was trying to create a simpler and more straight forward block when one just needs the SVGKImage and not the SVGKParseResult object. If you feel strongly about reusing the existing typedef, sure. 

Yes. I just noticed the wrong completion instance checking. The block is not invoked at all. *Sigh*

adamgit:
NB: I'm not saying the old block is perfect - the way we signal failure is
a bit opaque, could be made clearer (IIRC we return an SVGKParseResult with
a nil image, and inside the parseresult you can query for how many fatal
errors there were) - but it's a major part of our public interface at the
moment (it's what all user-code currently has to interface with) so I'd
like to keep using it, and/or modify it carefully.

teffi:
Noted. Will keep that in mind.

adamgit:
UPDATE: now we've started a 3.x branch ... it's a good time to improve / fix any interfaces. I don't think this one is easy :) - there's a lot of information to handle (the extra metadata needed by the parser, the image that's created (or null), and the parse-errors, and separately the parse-log), but now is a good time to come up with a cleaner alternative if you have one in mind.

PR #555, opened 2018-01-30
Disable and null cocoa lumberjack

alan-andrade:


alan-andrade:
whoops, this was intended to be on the personal fork. sorry.

PR #567, opened 2018-05-08
Carthage support

gbrixey:
I had to edit the framework search paths slightly so that the Carthage build would succeed.

If this PR is accepted, then after merging, a new tag should be pushed with a version higher than 2.0.0. The new tag would be the default version that Carthage would fetch if a user included SVGKit in their Cartfile without specifying a version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,8 @@ build
 
 #VScode
 .vscode
+
+# Carthage
+Carthage/Checkouts
+Carthage/Build
+*.zip
--- a/SVGKit-iOS.xcodeproj/project.pbxproj
+++ b/SVGKit-iOS.xcodeproj/project.pbxproj
@@ -2799,10 +2799,7 @@
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				ENABLE_TESTABILITY = YES;
-				FRAMEWORK_SEARCH_PATHS = (
-					"$(inherited)",
-					"$(PROJECT_DIR)/3rd-party-frameworks/CocoaLumberjack-2.2.0/iOS",
-				);
+				FRAMEWORK_SEARCH_PATHS = "$(PROJECT_DIR)/3rd-party-frameworks/CocoaLumberjack-2.2.0/iOS";
 				GCC_NO_COMMON_BLOCKS = YES;
 				GCC_PREFIX_HEADER = "SVGKitFrameworks/SVGKitFramework-iOS/SVGKitFramework-iOS-Prefix.pch";
 				GCC_WARN_ABOUT_RETURN_TYPE = YES_ERROR;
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -29,3 +29,8 @@ build
 
 #VScode
 .vscode
+
+# Carthage
+Carthage/Checkouts
+Carthage/Build
+*.zip
```
adamgit, line 36: Wouldn't this block all ZIP files everywhere, which would be a very suprising feature?

If Carthage has to store ZIP files somewhere in the project, then please prefix the path for it.

SVGKit-iOS.xcodeproj/project.pbxproj
```diff
@@ -2799,10 +2799,7 @@
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				ENABLE_TESTABILITY = YES;
-				FRAMEWORK_SEARCH_PATHS = (
-					"$(inherited)",
-					"$(PROJECT_DIR)/3rd-party-frameworks/CocoaLumberjack-2.2.0/iOS",
-				);
+				FRAMEWORK_SEARCH_PATHS = "$(PROJECT_DIR)/3rd-party-frameworks/CocoaLumberjack-2.2.0/iOS";
```
adamgit, line 2802: This breaks the project. Pleqse don't do this.

adamgit:
This alters the search paths (which isn't acceptable - check your Xcode setup, Xcode has had a lot of bugs with this in the past!).

Also, I'm concerned by the aggressive .gitignore changes.

SVGKit-iOS.xcodeproj/project.pbxproj
gbrixey: What exactly does it break? I can still build the frameworks and run the demo app with this change.

.gitignore
gbrixey: This line is only to ignore the output of `carthage archive`, and can be removed.

dreampiggy:
@gbrixey Could you remove that `FRAMEWORK_SEARCH_PATHS` modification. Use `Carthage carfile as dependency` because `CocoaLumberjack` supports Carthage as well. 

You can do it following these steps: 

1. Add this line in the project's cartfile.

```
github "CocoaLumberjack/CocoaLumberjack"
```

2. Then run `Carthage update`

3. In Xcode project, drag the built iOS framework into the target link framework. Only add `Carthage` folder into gitignore.

4. Finish. User who use carthage will automatically clone the `CocoaLumberjack` before install the SVGKit. It's no use to use the pre-built `CocoaLumberjack.framework` any more.

Maybe I can also create a PR for this work. I guess we can merge this support into upstream because this changes is more simple and suitable (Don't effect CocoaPods && Manually Install user)

SVGKit-iOS.xcodeproj/project.pbxproj
dreampiggy: It's better to add `$(inherited)`.
The project setting for `FRAMEWORK_SEARCH_PATHS` is wrong because it use `$(PROJECT_DIR)/3rd-party-frameworks/**`. Which may mass up the platform in different target. It's better to clear that and add each target for each platform value.

dreampiggy:
@gbrixey @adamgit Please see the modified version: https://github.com/SVGKit/SVGKit/pull/579
This one will not break current user (CocoaPods/Manually Install) and add Carthage support.

adamgit:
Now merged as #579

PR #571, opened 2018-07-04
Fix smooth-curve command (S/s and T/t) continuation behavior

pwesten:
The smooth curve commands (S/s and T/t) only continue with reflected control points from commands of the same type (C/c/S/s and Q/q/T/t, respectively), with prescribed behavior when following other types of commands--e.g. "move-to" or a curve of a different type--that requires the "current point" to be available. The SVGCurveZero sentinel was not sufficient for this purpose, as it omits that coordinate.

The included PNG reference image matches Safari, Chrome, and Xcode's own preview of the SVG file. This is how SVGKit rendered the new test before this fix:

m-to-s-to-t_before


pwesten:
@adamgit could I get your eyes on this PR? Thanks!

adamgit:
I thought I'd seen someone else approve this already - sorry!

PR #580, opened 2018-10-12
Update podspec to not pollute users xcconfig

dnkoutso:


adamgit:
I'm afraid I have no idea what this does -- can someone else who uses CocoaPods comment + Review please?

dreampiggy:
This one may be better. Because C++ standard library has many implementation (For example, GNU tools contains a `libstdc++`, however, clang tool contains a `libc++`). So there is one config called `CLANG_CXX_LIBRARY` in the Xcode to specify which one you want. (Even though Xcode 10 remove the `libstdc++` support...This works for old version :))

However, when using `libraries = 'c++'`, it will use the `default` one from user's project config (behave like a `$inherited`). For project using GNU tools, the `CLANG_CXX_LIBRARY` will be set to `libstdc++`. For project using clang tools, the `CLANG_CXX_LIBRARY` will be set to `libc++`. It's more intelligent than the hard-coded one.

dreampiggy:
@adamgit  And one more question about CocoaPods. I see there are a v2.0 release in the git tag and [github release page](https://github.com/SVGKit/SVGKit/releases/tag/2.0.0). However, the podspec files was not in the CocoaPods's [official source here](https://github.com/CocoaPods/Specs/tree/master/Specs/d/4/9/SVGKit). It shows the latest version is 1.2.0.

Because of this, I can not pass `pod lib lint` when my SVGCoder depedends on SVGKit's 2.x branch. Could you please at least upload the 2.0.0 podspec files into the source ?

You can do this, by using `pod` command line in the root of this project folder.

```bash
git checkout 2.0.0
pod trunk push --allow-warnings
```

---

OK...I found that the 2.0.0 tag's [podspec file](https://github.com/SVGKit/SVGKit/blob/2.0.0/SVGKit.podspec) contains issue about version. The version filed is set to `2.x` and the source filed is using a branch. So maybe we'd better not to upload this spec into trunk repo. We have to change it into the correct version and use the git tag in source filed.

adamgit:
The CocoaPods maintainers arbitrarily took a random version of SVGKit, didn't ask anyone, and then refused to let us replace/update it. This is a part of why we stopped supporting CP: the maintainers knowingly forced a buggy, broken, version of the library onto CP users and categorically refused to stop. We begged them not to.

If you can get them to delete SVGKit from their repository that would be awesome. No-one here wants to be responsible for an externally hosted unofficial copy that we don't control. 

If not, anything you can do to fix the cocoapods is fine, but not supported.

adamgit:
(PS: I don't believe anyone actually uses the cocoapods copy - that version is so old and buggy and has so many missing core features that it's practically useless)

dreampiggy:
@adamgit What about pushing a 2.0.0 tag into the CocoaPods's source ? I check and see you're the one of Owners. So you have right to push the new version into the CocoaPods trunk. Our [SDWebImageSVGCoder](https://github.com/SDWebImage/SDWebImageSVGCoder) want to pass the CI during `pod spec lint`. However, the old `1.2.0` version is broken to build on the current Xcode. Only the 2.x branch works.

```
➜  ~ pod trunk info SVGKit

SVGKit
    - Versions:
      - 0.0.1 (2014-05-19 21:43:42 UTC)
      - 1.0 (2014-05-19 21:49:54 UTC)
      - 1.2.0 (2014-07-01 11:48:47 UTC)
    - Owners:
      - Mirko Olsiewicz <mirko.olsiewicz@olsiewicz.de>
      - Alexandros Salazar <asalazar@weather.com>
```

adamgit:
I cannot do this. The last few times I tried their system crashed 
repeatedly and although they could have fixed it in minutes, they refused 
to help in any way. If I remember correctly they would not give access to 
their ssrver so even if we debuggex it we couldnt fix anyyhing ourselves. 
I'm not wasting any more time debugging their server.

adamgit:
To be clear: whatever cocoapods says, I believe they have never let me 
access their system (despite many attempts, emails, offers to prove 
identity etc - and for a long time they wouldn't accept any of us were 
maintainers on the project despite the very public, easily verifiable, info).

Collectively and individually we have put lots of time into trying to make 
cocopods work. The CP team have been unhelpful and damaging throughout. If 
you can get them to do anything at all, please do - everyone else has 
failed. Good luck!

adamgit:
To be clear: neither of the people listed on there are maintainers of svgkit.

Cocopods team knows this and repeatedly made clear they didnt care.

If you are using CP for production code, and believe that the CP repository 
is accurate or secure, I strongly advise you not to :). Experience has 
shown that it is smoke and mirrors and politics, not accurate nor sensibly 
maintained. After my experiences (and my colleagues experiences) we moved 
all our production projects off cocopaods - we have had noticeably fewer 
bugs since. Development is faster without CP.

dreampiggy:
@adamgit CocoaPods is just a central dependency manager. Not a devil. Maybe I guess you're not familiar with CocoaPods's command line ?

If you're not in the Owners of that list of SVGKit in CocoaPods trunk. You can claim that you're the owner of that repo. Goto https://trunk.cocoapods.org/claims/new .Then they can give back to your owner and you can push to trunk.

For CocoaPods configuration of project, I've updated the `podspec` files inside #581 PR. It works when using the branch dependency of `Podfile`.

adamgit:
We did all that. It didnt work. We asked CP to fix it. They wouldn't. We 
asked them to remove it when other  people claimed the project. They 
wouldn't. We asked them to edit the version tag since their system crashed 
whenever we tried to do it ourselves. They wouldn't.

... I didn't give up on it without trying :).

dreampiggy:
@dnkoutso Hi. I saw you're the contributor for CocoaPods dependency manager. Could you please help us to contact the CocoaPods server maintainer or other people,  to claim the repo for us ? Thanks in advance.

dnkoutso:
Hi there, apologies for the long delay here.

Yes I am a core contributor and while I do not know what has happened in the past I'd like to help here and understand the issues.

A lot of stuff has changed in CocoaPods.

This change is relatively harmless as the old DSL `s.xcconfig` was actually changing the `xcconfig` generated for the users project, not for the `xcconfig` generated fo SVGKit itself by CocoaPods.

`s.xcconfig` was deprecated because of that and was split between `s.user_target_xcconfig` and `s.pod_target_xcconfig`. The latter (with this change here) is the one used for your pod.

dnkoutso:
Let me also reach out with respect to the claim of the pod.

PR #581, opened 2018-10-13
Add macOS support in 2.x

dreampiggy:
@adamgit Hi. I'm the author of [SDWebImageSVGCoder](https://github.com/SDWebImage/SDWebImageSVGCoder), a SVG coder plugin for [SDWebImage](https://github.com/rs/SDWebImage) (A iOS image loading system). Since I want to support macOS (for simply usage, but not expert user), I found that the readme of this repo, point out to let user to use the [MaddTheSane/SVGKit](https://github.com/MaddTheSane/SVGKit/tree/master) fork.

However, his fork master branch is far away from current 2.x branch and it won't compile on Xcode 10. The author change too much of existing APIs and write some custom logic. So it's not easy to merge. So I decide to try it by myself, by taking minimal changes to current core code to bring it to macOS platform.

For macOS user, we can just simply treate `NSImage` as `UIImage`, `NSView` as `UIView`, and more, so current `UIImage` exporter can be also used for macOS. Then, I grab the MaddTheSane's `SVGKImageRep` class (which is one important for macOS but not iOS). After I fix all the compile issue, it seems the framework can work on macOS now.

No changes to current code on iOS/tvOS user. Just using the alias and fix some cross-platform code to make it work on macOS. You can see the diff. Some changes are related to project structure. For example, we re-defined the same `pch files, umbrella header files` for each target, however the content are nearlly same so I put them into single one to simplify the problem.

 If someone is interested in this, please review this. There may be some little issue (I run a simple demo on macOS and it works. But for large project it may not).


Source/AppKit additions/SVGKImageRep.h
```diff
@@ -0,0 +1,38 @@
+//
```
dreampiggy, line 1: This file is original from [MaddTheSane](https://github.com/MaddTheSane/SVGKit/blob/master/Source/AppKit%20additions/SVGKImageRep.h).

Source/DOM classes/Core DOM/Comment.h
```diff
@@ -11,7 +11,13 @@
 
 #import "CharacterData.h"
```
dreampiggy: Ugly fix. I found that MaddTheSane change all the class in `Core DOM` with the `SVGK` prefix, for example, `Attr -> SVGKAttr`, `Node -> SVGKNode`. I think it's a good idea and can solve many potential future issue. 

Since the user will not always directlly use class in `Core DOM`, I think renaming maybe adoptable. What do you think?

Source/DOM classes/SVG-DOM/SVGRect.h
```diff
@@ -38,3 +38,11 @@ CGRect CGRectFromSVGRect( SVGRect rect );
 
 /** Convenience method to convert to ObjectiveC's kind of size - ONLY the width and height of this rect */
 CGSize CGSizeFromSVGRect( SVGRect rect );
+
+NSString * _Nonnull NSStringFromSVGRect( SVGRect rect );
+
+#if SVGKIT_MAC
```
dreampiggy, line 45: Maybe these portable methods should be only used privately, but not in public API.

Source/UIKit additions/SVGKLayeredImageView.m
```diff
@@ -75,28 +79,33 @@ - (void)populateFromImage:(SVGKImage*) im
 	{
 #ifndef SVGK_DONT_USE_EMPTY_IMAGE_PLACEHOLDER
         SVGKitLogWarn(@"[%@] WARNING: you have initialized an [%@] with a blank image (nil). Possibly because you're using Storyboards or NIBs which Apple won't allow us to decorate. Make sure you assign an SVGKImage to the .image property!", [self class], [self class]);
-		
+#if SVGKIT_UIKIT
 		self.backgroundColor = [UIColor clearColor];
+#else
+        self.layer.backgroundColor = [NSColor clearColor].CGColor;
+#endif
```
dreampiggy: Moving these constant string into utils. Because now there are multiple files need this...No need to write twice.

Source/UIKit additions/SVGKLayeredImageView.m
```diff
@@ -75,28 +79,33 @@ - (void)populateFromImage:(SVGKImage*) im
...
         
-        NSString* svgStringDefaultContents = @"\
```
dreampiggy: Moving these constant string into utils. Because now there are multiple files need this...No need to write twice. See MaddTheSane's [code](https://github.com/MaddTheSane/SVGKit/blob/master/Source/BlankSVG.m)

SVGKitFrameworks/SVGKitFramework-iOS/SVGKitFramework-iOS-Prefix.pch
```diff
@@ -1,30 +0,0 @@
-//
```
adamgit, line 1: Why are all these .h and prefix headers deleted?

dreampiggy: Previouslly, each target create their own prefix header. However, the content are the same. So we can use a universal prefix header instead.

The same thing apply to the umbrella header (Which import all public headers), instead of copy files again and again. See the first commit.

Source/Exporters/SVGKExporterUIImage.m
```diff
@@ -14,14 +14,19 @@ +(UIImage*) exportAsUIImage:(SVGKImage*) image antiAliased:(BOOL) shouldAntialia
...
+#endif
+		SVGGraphicsBeginImageContextWithOptions( image.size, FALSE, scale);
+		CGContextRef context = SVGGraphicsGetCurrentContext();
 		
 		[image renderToContext:context antiAliased:shouldAntialias curveFlatnessFactor:multiplyFlatness interpolationQuality:interpolationQuality flipYaxis:FALSE];
 		
-		UIImage* result = UIGraphicsGetImageFromCurrentImageContext();
-		UIGraphicsEndImageContext();
+		UIImage* result = SVGGraphicsGetImageFromCurrentImageContext();
+		SVGGraphicsEndImageContext();
```
adamgit: We try to avoid using external utils classes. Originally the project used them, and it was a nightmare for anyone new to the project - e.g. this change here replaces "code that everyone knows and understands because it's standard iOS library" with "proprietary code that you have to go and open another file merely to discover that it's a facade for the code that - in fact - is merely standard code".

Source/Parsers/SVGKParseResult.h
```diff
@@ -17,8 +17,8 @@
 /** 0.0 = no parsing done yet, 0.x = partially parsed, 1.0 = parse complete (no fatal errors) */
 @property(nonatomic) double parseProgressFractionApproximate;
 
-@property(nonatomic,strong) SVGSVGElement* rootOfSVGTree; /*< both are needed, see spec */
-@property(nonatomic,strong) SVGDocument* parsedDocument; /*< both are needed, see spec */
+@property(nonatomic,strong) SVGSVGElement* rootOfSVGTree; /**< both are needed, see spec */
```
adamgit: All these changed comments are simply wrong? The official style - recognised by tools! - is "slash asterisk less than". Why these changes?

adamgit: (I double checked before adding the comment - it's /** for a comment that precedes a declaration, and /*< for a comment that postcedes a declaration. Or was I looking at the wrong places? Is there a new style that's superceded the classic Mac/obj-c commenting?)

Source/SVGKImage.h
```diff
@@ -137,6 +144,7 @@ typedef void (^SVGKImageAsynchronousLoadingDelegate)(SVGKImage* loadedImage, SVG
  */
 +(SVGKParser *) imageWithSource:(SVGKSource *)source onCompletion:(SVGKImageAsynchronousLoadingDelegate)blockCompleted;
 
+- (id)initWithContentsOfURL:(NSURL *)url;
```
adamgit: Is this method part of UIImage on OSX?

I have vague memory that this method is basically impossible to implement in SVG standard, because it provides too little information - hence the imageWithSource:: method, that provides extra metadata for resolving parsing issues.

Source/UIKit additions/SVGKFastImageView.h
```diff
@@ -40,14 +38,11 @@
...
 
-/** Apple has a bug in CALayer where their renderInContext: method does not respect Apple's own mask layers.
- 
- This is required to render SVGGradientElement's, and it is NOT a bug in SVGKit - it's in Apple's code. Until we
- can invent a workaround (or Apple fixes their bug), it's best to warn developers that their SVG will NOT render
- correctly
+/** The text implementation on OS X is different between CALayers and NSViews. If the CALayer is made in
+ SVGKLayeredImageView, it renders right-side up. Otherwise, it is upside-down.
  */
-+(BOOL) svgElementAndDescendentsHaveNoGradients:(SVGElement*) element;
```
adamgit: Why is this deleted?

Source/Parsers/SVGKParseResult.h
dreampiggy: Just Xcode warning. Because Xcode now parse all comment as [Doxygen](https://www.stack.nl/~dimitri/doxygen/manual/docblocks.html) format. (I don't know whether Xcode 6 or old version behavior, but when I become iOS developer I use Xcode 7).

The `<` is a doxygen syntax, means the comment should appear above the thing (such as property). So Xcode will warning you...Simple way is to use `/** < This comment is above property */`.


image

Source/UIKit additions/SVGKFastImageView.m
```diff
@@ -24,31 +24,35 @@ @implementation SVGKFastImageView
...
-			}
-				
-		}
-	}
-	
-	return TRUE;
+    return [self svgElementAndDescendents:image.DOMTree haveNoClass:[SVGTextElement class]];
+}
+
++ (BOOL)svgElementAndDescendents:(SVGElement*)element haveNoClass:(Class) theClass
```
adamgit: This seems a mess. This appears to be changing it to require pass in a parameter of a "class that the caller needs to know is a magic class that they MUST pass in the correct class (for no reason) because the code needs it or else the code breaks". WTF?

We have a FIVE LINE COMMENT in the original source explaining why this would be the wrong thing to do and is completely unnecessary --- unless the original comment is a lie, this change appears to ignore the comment completely, delete it, and then substitute meaningless illegible code for no purpose!

(I do not know if the original comment is true, or if it was true and Apple's implenmentation has now changed, or if it was never true. But for all three variations, this code change seems a bad one. If the original comment is no longer true, we can simplify the method name and delete the comment. But I don't see why anyone would want to extend the method, make it harder to use, and without explanation)

Source/UIKit additions/SVGKFastImageView.m
```diff
@@ -74,7 +78,12 @@ -(id)initWithFrame:(CGRect)frame
 	self = [super initWithFrame:frame];
 	if( self )
 	{
+#if SVGKIT_UIKIT
```
adamgit, line 75: What's the different between #if on SVGKIT_UIKIT and !(#if MAC_OS) ?

Source/UIKit additions/SVGKLayeredImageView.m
adamgit: This is tightly bound to the class it is in - it's data for the class implementation.

What are the other places that are using it?

Source/Exporters/SVGKExporterUIImage.m
dreampiggy: It's aim to keep macOS compatible. All the method call the `UIGraphics` method on iOS / tvOS without extra logic. However, macOS does not have this convience method (AppKit suck), you need to write long code to make it work (See the implementation).

Currently seems only one code use the method. So I can remove that from public API and put the implementation here as private API.

adamgit:
Overall looks great.

I have a few concerns (hilighted in the comments).

Source/SVGKImage.h
dreampiggy: Emm....Because [MaddTheSane's fork](https://github.com/MaddTheSane/SVGKit/blob/master/Source/SVGKImage.h#L133) use this method. And his [SVGImageRep implementation](https://github.com/MaddTheSane/SVGKit/blob/master/Source/AppKit%20additions/SVGKImageRep.m#L130) need this API.

I think maybe this `URL` is just fileURL, but not network URL. Maybe I can remove the public API, use file path instead.

Source/DOM classes/Core DOM/Comment.h
adamgit: No-one has invented a satisfactory solution to this problem - ObjectiveC is too weak a language to solve it neatly.

I believe the best solution is:

1. Create a new open-source project "CoreDOM"
2. Move all the DOM source code to that project
3. Add the CoreDOM project as one of SVGKit's dependencies

...then anyone is free to re-implement the public headers of CoreDOM and replace it.

(everything else people have suggested or thought of and tried ... hasn't worked. ObjectiveC just doesn't support dynamic linking, late binding, replaceable vendor code, etc)

NB: there may be a reason that even this solution doesn't work -- I can't remember if anyone tried it :) -- but I believe the only reason no-one did it was that it felt like a lot of hassle to setup an extra project in github?

adamgit: MaddTheSane's solution is really not good. It prevents anyone from ever replacing the CoreDOM implementation (most of which I hacked out in a few days, just implementing the spec as simply as possible, and probably could be greatly optimized / expanded)

Source/Parsers/SVGKParseResult.h
adamgit: Ah, OK, so it looks like the standard has changed :). Fine.

Source/UIKit additions/SVGKFastImageView.m
dreampiggy: I just copy the comment from [MaddTheSane's fork - SVGKFastImageView](https://github.com/MaddTheSane/SVGKit/blob/master/Source/UIKit%20additions/SVGKFastImageView.h#L66-L68). Maybe he know the reason Apple's layer implementation on macOS need this ?

I can try to delete this and see the result.

Source/Exporters/SVGKExporterUIImage.m
adamgit: Having looked at the utils class now, I think I'd prefer to see:

#if IPHONE
 UIGraphics....
#else
 SVGKGraphics...
#endif

NB:
1. the code path for iOS is clear and obvious
2. the caller still only has a simple case - one line for either OS
3. SVGK instead of SVG (SVG is reserved by the SVG spec)

Source/UIKit additions/SVGKFastImageView.m
dreampiggy: Current seems have the same meaning. Because we support only iOS/tvOS/macOS. `SVGKIT_UIKIT` is stand for iOS or tvOS. So `#if !SVGKIT_UIKIT` is actually means macOS.

However, maybe someone in the future want to add watchOS support ? This assume will fail.

Source/SVGKImage.h
adamgit: Yep, be careful to check that one. SVG is actually specified differnetly by W3C for whether it's loaded from a web URL, or if the URL is being used as a file:// (I think! It's been a long time since I read this).

Source/Exporters/SVGKExporterUIImage.m
dreampiggy: I'v remove the API from public into private. Renaming is OK.

Source/UIKit additions/SVGKFastImageView.m
adamgit: This patch uses both, and unless there's a strong reason for that, I think it should only use one style or the other.

Does watchOS have UIKit? Or is it an incompatible UIKit?

I like the idea of using a SVGKIT_UIKIT everywhere, as that seems to be accurate (the problem isn't "OSX" so much as "OSX currently doesn't have a UIKit implemetnation built-in", right)?

Source/Parsers/SVGKParseResult.h
dreampiggy: Actually...You can just merge this PR: #578  He do the same thing as me...

I'll rebase local branch after that PR merged. To simplify diff changes.

SVGKitFrameworks/SVGKitFramework-iOS/SVGKitFramework-iOS-Prefix.pch
adamgit: Ah, right. I have a suspicion that there was some reason for needing the separate headers - some bug or "feature" in xcode versions, or some problem to do with the build-for-app-store phase, where it got confused if you used a shared header.

I might be being paranoid here ;), but unless someone can test this change with a push-to-asset-store release, I'd prefer to shift it to a separate patch, and do that on a side branch and have people use it and wait until soneone's confirmed they got iphone, ipad, etc builds to go fine.

(i.e. don't let it slow down / poison the rest of this patch - if it's merely a code-cleanup thing?)

Source/UIKit additions/SVGKFastImageView.h
dreampiggy: Ops..I think this API is un-used. Because here is one `svgImageHasNoGradients`.

Anyway, this may break old API (at least we should deprecate but not remove it). I will keep this.

dreampiggy: One question, should't these API be wrapped inside `TEMPORARY_WARNING_FOR_APPLES_BROKEN_RENDERINCONTEXT_METHOD` macro ?

If someday we change the .m files value to 0, the public header still contains this API (but no implementation actually), it may crash because of unrecognized selector.

Or we can remove the `TEMPORARY_WARNING_FOR_APPLES_BROKEN_RENDERINCONTEXT_METHOD` macro inside implementation for these API. So that this API is always available and don't cause issue when we change the value.

adamgit:
If it's caused by the same issue, then what you say sounds correct to me.

(or maybe it needs a new macro, if it's a similar but different root cause :))

Source/UIKit additions/SVGKFastImageView.m
dreampiggy: Correct. OS X use AppKit, iOS / tvOS use UIKit. At least now. Apple may release a real UIKit for macOS in the future (next year), but however, there are no API guideline for that so current we have no way to know how to detect. If they release, I can also try to update the way to support.

WatchOS contains a small subset of UIKit, actually the `UIKitCore.framework`, the only public class are `UIImage`, `UIFont`, `UIColor`. All the `UIView`/`UIViewController/UIResponder` class is not available for watchOS.

adamgit:
So ... make our macro:

#define SVGK_UIKIT_FULL
#define SVGK_UIKIT_MINIMAL

with iOS getting both, watchOS getting only the second, and  OS X getting none?

Source/UIKit additions/SVGKLayeredImageView.m
dreampiggy: The [MaddTheSane's SVGKImageRep](https://github.com/MaddTheSane/SVGKit/blob/master/Source/AppKit%20additions/SVGKImageRep.m#L209)'s default init use this value... 😅 

Or maybe we can just ignore the default init method ?

Source/DOM classes/SVG-DOM/SVGRect.h
MaddTheSane: An alternative is to use preprocessor macros to redefine, i.e.  NSStringFromCGRect to NSStringFromRect:

````
#define NSStringFromCGRect NSStringFromRect
````

Although I agree that it could be put in a better place.

File: Source/SVGKImage.h (hunk 9 above)
Comment (MaddTheSane) [reply Source/SVGKImage.h]: Yeah, it was supposed to be for file URLs, but it could be used for network URLs as well. A disclaimer saying that using it for network URLs might be wise.

File: Source/DOM classes/Core DOM/Comment.h (hunk 2 above)
Comment (dreampiggy) [reply Source/DOM classes/Core DOM/Comment.h]: @adamgit You means some advanced user, may provide a custom implementation of DOM object used in our repo ?

Maybe I think the better way, it to define all of DOM object, with the same name protocol in Objective-C. For example, we can define `@protocol Document @end` as well as `@interface Document<Document> @end`. Then people who use the custom implementation can replace the class of we used by conforming to the protocol.

However, it's a big changes. So I think they can just fork the repo and use another same class name, but changing to different implementation instead. If they'd like this approach, I think move the Core DOM code into another repo is acceptable. Because they can focus on DOM spec but not SVG parser or something non-related thing.

File: Source/DOM classes/Unported or Partial DOM/SVGElement.m [hunk 13]
```diff
@@ -254,9 +254,6 @@ - (void)postProcessAttributesAddingErrorsTo:(SVGKParseResult *)parseResult  {
                 value = [self getAttribute:@"gradientTransform"];
             }
 		
-#if !(TARGET_OS_IPHONE) && ( !defined( __MAC_10_7 ) || __MAC_OS_X_VERSION_MIN_REQUIRED < __MAC_10_6_7 )
```
dreampiggy, line 257: The macOS min-deployment target version is set to macOS 10.10 (To let NSImage exporter works). So this check is useless.

Source/SVGKImage.h
dreampiggy: Keep this. Because currentlly we already have one API `+[SVGKImage imageWithContentsOfURL:]`, while we don't export the `-[SVGKImage initWithContentsOfURL:]` this API. 

However, from the naming, these two should be either available or not at the same time. So I just export the `-[SVGKImage initWithContentsOfURL:]` one. Keep @MaddTheSane code.

Source/UIKit additions/SVGKFastImageView.m
dreampiggy: @MaddTheSane What's the problem that macOS's `SVGFastImageView` rendering `<text>` element ? If it's only something about quartz coordinate system (non-flip on macOS,flip on iOS), it's easy to fix by using [NSView.isFlipped](https://developer.apple.com/documentation/appkit/nsview/1483532-flipped)

dreampiggy:
@adamgit I remove all `TARGET_OS_IPHONE` with the correct `SVGKIT_UIKIT` check. Because `TARGET_OS_IPHONE` actually means (iOS or iPhone Simulator or Apple TV or Apple Watch).

`SVGK_UIKIT_FULL` currentlly seems no use, only if we want to add watchOS support we can define like that. (Or using `#if SVGKIT_WATCH || SVGKIT_UIKTI`). But however, it seems that all QuartZ method like `CALayer` is not available on watchOS. So the main feature of our framework is broken (We can not render SVG!). Until Apple introduce the `CALayer` on watchOS, we can not port it into that platform 😢

dreampiggy:
@adamgit Also, because you say that `SVGKGraphicsBeginImageContenxt` this one is not standard for our developer. So I seperate the code into `SVGKExporterNSImage` new class. On iOS && tvOS we use `SVGKExporterUIImage`, on macOS we use `SVGKExporterNSImage`. Better on naming.

Source/UIKit additions/SVGKFastImageView.m
MaddTheSane: IIRC, if rendered using CoreAnimation, text would be right side up. But if rendered without CoreAnimation, it would be upside-down.

dreampiggy: @MaddTheSane You mean, if I use `SVGFastImageView` as layer-backed view (wantsLayer = YES), the render result is correct ? But if I don't use layer (wantsLayer = NO), the render is flipped ? Sounds strange.

I'll try with the branch and using a demo. If so, I guess it's not hard to fix. (For example, we can change something in the `SVGKLayer` parsing and add a transform specially for macOS)

dreampiggy: I found it's a bug because of the implementation of `SVGTextElement`. I'll try to fix it on macOS

dreampiggy: @MaddTheSane My fault. It seems the Apple's bug in `-[CALayer renderInContext:]` for `CATextLayer`. Best way to solve it, it to turn on `USE_SUBLAYERS_INSTEAD_OF_BLIT = 1`. Which will use `sublayers` and it works like `SVGKLayeredImageView`.

---

If you still want to use it as normal drawRect view. Go on and see below:

It seems that `CATextLayer` render in the context will flip the content (Text String) on macOS. However, from QuartZ coordinate system programming, it should use the `non-flipped` one (seems bug or Apple's internal hack). One way to solve this, it to override the `contentsAreFlipped` property and return YES. However, it needs a subclass like `CATextLayerBugFixForMac`.

Another hack way, we can use the `contentsTransform` of CALayer (Used by `NSImageView` when rendering JPEG images, which EXIF orientation is upside down). Here is the simple way of code.

```objectivec
// SVGKFastImageView.m

            CALayer *layerTree = self.image.CALayerTree;
            [self preprocessLayerTree:layerTree];
	    [layerTree renderInContext:context];
            [self postProcessLayerTree:layerTree];


- (void)preprocessLayerTree:(CALayer *)layerTree {
#if SVGKIT_MAC
    if ([layerTree isKindOfClass:[CATextLayer class]]) {
        // Hack to apply flip for text content
        NSAffineTransform *flip = [NSAffineTransform transform];
        [flip translateXBy:0 yBy:layerTree.bounds.size.height];
        [flip scaleXBy:1.0 yBy:-1.0];
        [layerTree setValue:flip forKey:@"contentsTransform"];
    }
#endif
    for (CALayer *layer in layerTree.sublayers) {
        [self preprocessLayerTree:layer];
    }
}

- (void)postProcessLayerTree:(CALayer *)layerTree {
#if SVGKIT_MAC
    if ([layerTree isKindOfClass:[CATextLayer class]]) {
        // Hack to recover flip for text content
        NSAffineTransform *flip = [NSAffineTransform transform];
        [layerTree setValue:flip forKey:@"contentsTransform"];
    }
#endif
    for (CALayer *layer in layerTree.sublayers) {
        [self postProcessLayerTree:layer];
    }
}
```

---

OK... I found that images are flipped as well. Good Apple, so we fix them together... See my commit about it. (`SVGKImage+CGContext.m`)

dreampiggy:
I port the @MaddTheSane 's fork branch MacOS demo to the 2.x. It works well. And after that fix mentioned in the comment, now `<text>` or `` on macOS using `SVGKFastImageView` can render correctly without flip. 

image

Source/SVGKImage.h
adamgit: Looking at the source, initWithContentsOfURL is implemented safely at the moment (it constructs an SVGKSource and fills it out with the URL before proceeding).

Unless I've misread that, I see no reason to avoid having both methods.

(I think there was a time when the + method worked better, because it did some housekeeping, but the - method did not)

SVGKitLibrary/Demo-OSX/SKSVGObject.h
```diff
@@ -0,0 +1,40 @@
+//
+//  SKSVGObject.h
```
adamgit: The name of this looks odd. Is it  trying to do a two letter naming prefix? Should it be SVGK instead?

(since it's in the demo I guess it doesn't hugely matter, but if it's meant to be following the naming scheme, should probably follow the same one as the rest of the library)

SVGKitLibrary/Demo-OSX/en.lproj/Credits.rtf
```diff
@@ -0,0 +1,29 @@
+{\rtf0\ansi{\fonttbl\f0\fswiss Helvetica;}
```
adamgit: Either delete this file or have it direct people straight to the github repository URL?

Source/DOM classes/Core DOM/Comment.h
adamgit: TL;DR: unless someone moves the DOM implementation into a new repo, I prefer leaving it with the current names (that were defined by the official W3C standard! They are supposed to be the same in every platform, every programming language. That's explicitly how the spec was written - and is why they look strange on some platforms)

Source/Exporters/SVGKImage+CGContext.m
```diff
@@ -95,4 +98,47 @@ -(void) renderToContext:(CGContextRef) context antiAliased:(BOOL) shouldAntialia
 	SVGKitLogVerbose(@"[%@] renderToContext: time taken to render CALayers to CGContext (perf improvements:%@): %2.3f seconds)", [self class], perfImprovements, -1.0f * [startTime timeIntervalSinceNow] );
 }
 
+- (void)preprocessRenderingLayerTree:(CALayer *)layerTree {
```
adamgit: Good idea, but please rename the two methods to make it clear that they are NOT an intentional feature of the class, but rather a temporary workaround to deal with one-off rendering issue in Apple code.

(that way they won't get used by people in future thinking that this was the intended class structure).

e.g. something like:

-(void)temporaryWorkaroundPreprocessRenderingLayerTree:


.. AND: copy/paste the // comments up into a doxygen comment for the method declaration, as they explain why the method has to exist.

Source/SVGKDefine.h
```diff
@@ -0,0 +1,62 @@
+//
+//  SVGKDefine.h
+//  SVGKit-iOS
+//
+//  Created by lizhuoli on 2018/10/14.
+//  Copyright © 2018年 na. All rights reserved.
```
adamgit: I'd like a sentence or several here explaining why this header exists.

I think it's "a central place to hold lots of project-wide defines that make it easier to write clean code that works across the different Apple platforms, which are sometimes incompatible with each other" ... or something like that! ... but I'm not sure.

Source/UIKit additions/SVGKFastImageView.m
```diff
@@ -6,6 +6,9 @@
 #import "SVGGradientElement.h" 
 #endif
 
+#define SetupLayerBackedView() \
+self.wantsLayer = YES;
```
adamgit: Is there a particular reason to make this a define, instead of just inlining it into the place where it's used?

Source/UIKit additions/SVGKLayeredImageView.m
```diff
@@ -4,6 +4,10 @@
 
 #import "SVGKSourceString.h"
 
+#define SetupLayerHostingView() \
```
adamgit: Again - any reason not to inline this into the place its used?

Source/UIKit additions/SVGKFastImageView.m
MaddTheSane: None that I can think of.

Source/UIKit additions/SVGKLayeredImageView.m
MaddTheSane: Again, none that I can think of.

Source/UIKit additions/SVGKFastImageView.m
dreampiggy: I can just remove these into inline code...Because I there are one `SetupLayerHostingView` macro in `SVGKLayeredView`.

Source/Exporters/SVGKImage+CGContext.m
dreampiggy: Done.

Source/SVGKDefine.h
dreampiggy: Done.

Source/UIKit additions/SVGKFastImageView.m
dreampiggy: Done.

Source/UIKit additions/SVGKLayeredImageView.m
dreampiggy: @MaddTheSane The code is from your fork branch actually 😅. Done.

SVGKitLibrary/Demo-OSX/SKSVGObject.h
dreampiggy: Actually all of these demo code is from @MaddTheSane 's work so I just move entire files. See his demo here: [MaddTheSane-SKSVGObject.h](https://github.com/MaddTheSane/SVGKit/blob/master/XCodeProjectData/Demo-OSX/SKSVGObject.h)

I can rename it to `SVGKObject`, but I think it's just a lightweight model to represent the SVG test to make it work for tableView. Is that really necessary ?

SVGKitLibrary/Demo-OSX/en.lproj/Credits.rtf
dreampiggy: Removed.

Source/UIKit additions/SVGKLayeredImageView.m
MaddTheSane: I know. I can't remember why I did it that way. Maybe it was copy-pasted from something?

Source/SVGKImage.h
dreampiggy: @adamgit What means `initWithContentsOfURL is implemented safely at the moment`.

The differens between `imageWithContentsOfURL:` and `initWithContentsOfURL:` is because of Objective-C's naming conventions before ARC was introduced. Methods with `init*` should create a object with +1 retain count. Methods with `xxxWith*`  should create a autoreleasing object with +0 retain count. See all API provided by apple like [-[NSArray array]](https://developer.apple.com/documentation/foundation/nsarray/1460120-array?language=objc) && [-[NSArray init]](https://developer.apple.com/documentation/foundation/nsarray/1414315-init?language=objc).

However, when ARC was introduced, these assume actually take no effect. Because ARC will insert code about retain && release. The autoreleasing object should now be explit set using `@autoreleasing` pool. However, these naming was kept because of large code base. And a class method like `+[xxx imageWithContentsOfURL:]` is just a simply call with `[[xxx alloc] initWithContentsOfURL:]`.

The `@synchronized` here is actually take no effect. I read all function call of this, there are no thread-safe issue to make us using a lock on the object level. The only potential issue is that `globalSVGKImageCache`, a mutable dictionary without any synchonize method. If we want to keep it thread-safe, use `lock` just before any reading-writing operation to that dictionary, or use the thread-safe class [NSCache](https://developer.apple.com/documentation/foundation/nscache). It's no need to put un-related method into `@synchronized`.

adamgit:
I didn't look at the sync issues at all - I assumed that  was unchanged 
(and, yes, it's a serious issue! These methods are frequently used from 
different threads).

What I meant was that at one point in the past we had a partially incorrect 
implementation of one of these methods, I think. It worked fine until you 
tried to load an SVG with html/css/document  relative links inside it.

That's the only reason I could think of why one would be in the interface 
and the other wouldn't.

adamgit:
Id probably rename it to take out the prefix entirely :)

dreampiggy:
> I didn't look at the sync issues at all - I assumed that was unchanged (and, yes, it's a serious issue! These methods are frequently used from different threads). What I meant was that at one point in the past we had a partially incorrect implementation of one of these methods, I think. It worked fine until you tried to load an SVG with html/css/document relative links inside it. That's the only reason I could think of why one would be in the interface and the other wouldn't.

Is this reply to that `imageWithContentOfURL:` usage ?

I means, since there are already one `imageWithContentOfURL:` in the public API, and the correspond `initWithContentsOfURL:` exist, so we can expose it as well. And it's also what @MaddTheSane does.

Anyway, if you're careful about chaning existing codebase, I can simply to use `imageWithContentOfURL:` instead. They take the same effect actually.

SVGKitLibrary/Demo-OSX/SKSVGObject.h
dreampiggy: All class in Objective-C should use prefix. Beucase Objective-C is strict super set of C. Which does not have namespace like C++/Java/Swift or any modern language. If someone name a class, anyone else can not name it the same, which will cause naming conflict.

So modern Objective-C toolchain, always use prefix, even for demo project. I think @MaddTheSane create the OS X demo with `SK` prefix, so this is why his model called `SKSVGObject`...

adamgit: Firstly: only Apple is meant to use 2-letter prefixes, 3 or 4 are meant to be the minimum for 3rd party frameworks (even using 4-letters isn't enough - many open-source projects already get collisions. ObjC wasn't designed to be used in public! :)).

Secondly: a class with a prefix is declaring itself to be part of the framework. Our demo classes are not part of the framework - they are standalone demos that could easily be deleted, or shipped separately. If a class in the demo is so useful that it might be worth including in the main framework, then it could have a prefix (although we ought to move such classes out of the demo and into the main source tree somewhere - I think SVGKLayeredView is frequently used by people, although it was originally just a quick and dirty hack I wrote).

Briefly: if a header file is included in the prefix header (i.e. it's part of the API that SVGkit is exporting to the world), it should be SVGK* ... if it's not, then it shouldn't.

I'd be OK with a different prefix - e.g. SVGKITDEMO - that makes it clear these classses are all NOT part of the core distribution.

dreampiggy: @adamgit So you mean we need to change the prefix of demo ?

Currentlly they use `SKSVG` for prefix, isn't it enough ? The core framework in SVGKit does not use it at all.

Anyway, if you can provide another prefix, I just do a renaming. My aim is to merge the macOS support in the upstream so that I don't need to let our SDWebImage users to use a fork repo for dependency instead.

adamgit:
This now shows merge conflicts with the Radial/Gradient PR I merged yesterday.

Maybe needs you to rebase it on the latest?

dreampiggy:
@adamgit What about I merge the latest 2.x to this local branch. It seems using rebase to local will cause many conflict. (Because many suck `xcode.pbxproj` changes) ?

dreampiggy:
@adamgit What about I merge the latest 2.x to this local branch. It seems using rebase to local will cause many conflict. (Because many suck `xcode.pbxproj` changes) ?

dreampiggy:
@adamgit What about I merge the latest 2.x to this local branch. It seems using rebase to local will cause many conflict. (Because many suck `xcode.pbxproj` changes) ?

dreampiggy:
@adamgit What about I merge the latest 2.x to this local branch. It seems using rebase to local will cause many conflict. (Because many suck `xcode.pbxproj` changes) ?

Or maybe I need to square some commit (expecially some adding and deleting to the same files) into one to simplify rebase process.

dreampiggy:
@adamgit What about I merge the latest 2.x to this local branch. It seems using rebase to local will cause many conflict. (Because many suck `xcode.pbxproj` changes) ?

Or maybe I need to square some commit (expecially some adding and deleting to the same files) into one to simplify rebase process.

dreampiggy:
@adamgit What about I merge the latest 2.x to this local branch. It seems using rebase to local will cause many conflict. (Because many suck `xcode.pbxproj` changes) ?

Or maybe I need to square some commit (expecially some adding and deleting to the same files) into one to simplify rebase process.

dreampiggy:
@adamgit What about I merge the latest 2.x to this local branch. It seems using rebase to local will cause many conflict. (Because many suck `xcode.pbxproj` changes) ?

Or maybe I need to square some commit (expecially some adding and deleting to the same files) into one to simplify rebase process.

dreampiggy:
@adamgit What about I merge the latest 2.x to this local branch instead of rebase. It seems using rebase to local will cause many conflict. (Because many suck `xcode.pbxproj` changes) ?

Or maybe I need to square some commit (expecially some adding and deleting to the same files) into one to simplify rebase process.

dreampiggy:
@adamgit What about I merge the latest 2.x to this local branch instead of rebase. It seems using rebase to local will cause many conflict. (Because many suck `xcode.pbxproj` changes) ?

Or maybe I need to square some commit (expecially some adding and deleting to the same files) into one to simplify rebase process.

dreampiggy:
GitHub seems down :) I've rebase and update again. The new bugfix for radial gradient && linear gradient works as well on macOS. However, for linear gradient, since Apple does not contains bug render the `CAGradientLayer` like it on iOS, maybe we can provide a macro or something to let user use the super method for render instead ?

Demo-iOS.xcodeproj/project.pbxproj
```diff
@@ -1043,7 +1075,7 @@
 					"../../iOS/**",
 					"../../Core/**",
 				);
-				IPHONEOS_DEPLOYMENT_TARGET = 5.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
```
adamgit, line 1078: I'm fine with making iOS8 the minimum, but not in the 2.x branch. That's a big jump, and we should push a tagged release first, and move to e.g. a 3.x branch.

SVGKit.podspec
```diff
@@ -2,7 +2,9 @@ Pod::Spec.new do |s|
   s.name        = 'SVGKit'
   s.version     = '2.0.1'
   s.license     = 'MIT'
-  s.platform    = :ios, '5.0'
+  s.osx.deployment_target = '10.10'
+  s.ios.deployment_target = '5.0'
```
adamgit, line 6: This is out of sync with the ios 8.0 min above?

SVGKit.podspec
```diff
@@ -14,12 +16,14 @@ Pod::Spec.new do |s|
                  'Moritz Pfeiffer' => 'moritz.pfeiffer@alp-phone.ch' }
   s.source   = { :git => 'https://github.com/SVGKit/SVGKit.git', :branch => "2.x" }
 
-  s.ios.source_files = 'Source/*{.h,m}', 'Source/DOM classes/**/*.{h,m}', 'Source/Exporters/*.{h,m}', 'Source/Parsers/**/*.{h,m}', 'Source/QuartzCore additions/**/*.{h,m}', 'Source/Sources/**/*.{h,m}', 'Source/UIKit additions/**/*.{h,m}', 'Source/Unsorted/**/*.{h,m}'
-  s.tvos.deployment_target = '9.0'
+  s.source_files = 'Source/*{.h,m}', 'Source/**/*{.h,m}'
+  s.ios.exclude_files = 'Source/AppKit additions/*{.h,m}', 'Source/Exporters/SVGKExporterNSImage{.h,m}'
```
adamgit, line 20: What impact does it have in CP to exclude these files?

SVGKitLibrary/Demo-OSX/SKSVGObject.h
adamgit: My preference is for ALL demo classes to have NO prefix. They are NOT part of the library,  therefore there is no reason for them to have a prefix!

(but also: SVGK*ImageView classes are clearly part of the library by now - they should be moved out of the Demo folder and into some other folder - maybe the folder for UIKit-specific additions?)

adamgit (changes requested):

Demo-iOS.xcodeproj/project.pbxproj
dreampiggy: This is demo project, not the framework project...What's the problem cause min deployment target version ?

SVGKit.podspec
dreampiggy: ? This is framework target version, iOS 5 is from the SVGKit-iOS.xcproj min deployment target version. Mac OS X support is new created by me, not available before. So this does not obey any [sem version](https://semver.org/) constraint.

SVGKit.podspec
dreampiggy: This files does not availble for iOS, it's specially for Mac OS X.

CocoaPods's podspec files is a build workflow base on the expression. You express to tell which files are compiled. It's not like the GNU makefile (Which is based on the step, you write code to till what step to do).

Our Xcode project SVGKit-iOS.xcproj have the same configuration. This files are excluded from iOS/tvOS.

dreampiggy: See the configuration in the Xcode project.

image

image


Actually, CocoaPods does not need you to configuration anything in the Xcode and ignore them at all. All the thing is done using their syntax and human-readable expression. However, for Carthage user to manual installed user, they need Xcode configuration. So both of these two are used in the large project like our team's framework SDWebImage/CocoaLumberjack. (There are also tools to automatically sync the Xcode configuration to CocoaPods podspec)

dreampiggy:
For `SKSVGObject`, this class's prefix is not anything we used in the framework. We use `SVG` && `SVGK` prefix. But not this `SKSVG`. So this may not cause issue.

Actually, modern Objective-C Programming Guide, poinit one rule, which most of my team member, my company's colleague agree: **ALL THE CLASS** should have prefix, wherever you write code on the project, or on the framework.

Because Objective-C is suck on namespace, does not allow any duplicated symbol at compile time/runtime (The historical reason from C). Unlike Swift/Rust/C++/Go or any other modern language, they using namespace to allow the same symbol exist, the namespace can be defined in code or implict from the project name.

I guess maybe recently you're not writing too much of Objective-C code, or it's your team's coding style. But anyway, my aim is to merge this into upstream. If you prefer this coding style, I can rename all of code in the Demo to use without any prefix.

SVGKit.podspec
dreampiggy: If you think this Mac OS X min deployment target version is higher. (From our company's data, Mac OS X 10.10 and before contains less than 10% total global Mac market). I can actually make it down to Mac OS X 10.9, using some compatible code.

For iOS min deployment tartget version. I didn't change the value. However, iOS 5 built's App actually does not allowed to submitted to App Store. See: https://developer.apple.com/news/?id=05072018a

So I guess this iOS min deployment target version is just a hint and does not cause actual result. People have to use Xcode 9 to submit to App Store or will be rejected. And Xcode 9 does not allows to to set min deployment target version to iOS 7 or before. (Automatically correct it to iOS 8)

dreampiggy:
@adamgit Please see the updated commit and comment reply. Any extra review issue ?

Demo-iOS.xcodeproj/project.pbxproj
adamgit: Ah! Sorry, reading the huge patch I didn't notice that was the demo project - my mistake.

Yep, I have no objection to the demo project being iOS8+ :). Anyone who's making new projects today and targetting less than 8 is ... um ... probably not someone we would support anyway (too expensive/difficult).

SVGKit.podspec
adamgit: It was the iOS version I was worrying about, because I misunderstood which xcodeproj file I was looking at.

Good point re: Xcode blocking setting the deployment target lower. What I would like to do is get this huge patch approved, merged, ... and then declare 2.x complete, push a tagged version, and move to a 3.x branch, where we can throw away everything that's less than - say - iOS 9.

SVGKit.podspec
adamgit: Sorry, I should have been explicit - it was the "AppKit additions" I was querying on that line?

adamgit:
> Actually, modern Objective-C Programming Guide, poinit one rule, which most of my team member, my company's colleague agree: **ALL THE CLASS** should have prefix, wherever you write code on the project, or on the framework.
> 

Yes, this is a good approach.

BUT ... the point here is that we *do not expect anyone to use the Demo code in their own projects*. The demo's are provided as tutorials, examples on how to use the framework. We include them in the project for convenience, so that new programmers don't have to google to find usage / documentation.

I want to draw a clean line between "code in the framework, that we support -- and we NEVER break it" and "code in the demo projects, that is provided as-is, with no support".

> I guess maybe recently you're not writing too much of Objective-C code, or it's your team's coding style. But anyway, my aim is to merge this into upstream. If you prefer this coding style, I can rename all of code in the Demo to use without any prefix.

Most frameworks I use day-to-day do *NOT* namespace their tutorials. Tutorial code is application code - if you use the tutorial code in your own project, you're expected to move it into your own namespace.

adamgit:
Final note: I don't like the way our demo projects (both the existing iOS demo, and the new Mac demo) pollute the namespaces. But we should probably postpone that decision to a future patch.

I don't see any blockers here. I would *like* to see *both* the Demo projects have their classes renamed. It is confusing and messy at the moment. But I'm happy to fix that later, in the interests of getting this merged!

adamgit (approved):

SVGKit.podspec
onnoweb: This change seems to break the build for me on iOS. SVGKExporterNSImage.h is excluded but SVGKExporterNSImage.m is not.

adamgit: Ouch. Thanks for spotting that. We're doing a final check of the build on all projects now - hopefully @dreampiggy can correct that (with a new PR / patch).

dvanwinkle:
This commit broke iOS builds.

dreampiggy:
@dvanwinkle Could please provide more information about how iOS builds break. ? I try to build the framework, it works.

See the output: 

```basb
➜  SVGKit git:(2.x) xcodebuild build -project SVGKit-iOS.xcodeproj -scheme 'SVGKitFramework-iOS' -sdk iphonesimulator -configuration Debug | xcpretty -c
▸ Building SVGKit-iOS/SVGKitFramework-iOS [Debug]
▸ Check Dependencies
▸ Processing Info.plist
......
▸ Signing /Users/lizhuoli/Library/Developer/Xcode/DerivedData/SVGKit-iOS-blqbsxtvbyporfauofvqlurdipfh/Build/Products/Debug-iphonesimulator/SVGKit.framework
▸ Build Succeeded
```

If you means not the built result, but about header import usage. (For example, Objective-C user) It may because of something you use case inside your project, for example, you use SVGKit which does not include the umbrella header.

```objective-c
#import <SVGKit/SVGKit.h> // correct, use umbrella headers, like <UIKit/UIKit.h>
@import SVGKit; // correct, better syntax
// don't do
#import <SVGKit/SVGKFastImageView.h> // wrong
#import <SVGKit.h> // wrong
#import "SVGKit/SVGKit.h" // wrong
#import "SVGKit.h" // wrong
```

This may cause issue because some public header use the prefix macro contains `SVGKIT_UIKIT`, which is defined in a pch file of the project, but not used inside each seperated headers.

dreampiggy:
@dvanwinkle OK...I found the typo inside `podspec` files. Which use the wrong syntax. Fix it now.
See #594 

Also, I create a PR to support user who don't use umbrella headers `SVGKit.h`, import the cross-platform define headers just before those public headers which use `SVGKIT_UIKIT` or `SVGKIT_MAC`.

PR #584, opened 2018-10-19
Fix the implementation of radial gradient and linear gradient

dreampiggy:
# Summary

This is a render fix, for both [radialGradient](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/radialGradient) and [linearGradient](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/linearGradient) to match most of SVG files rendering on WebKit browser like Chrome or Safari.

# Current issue

The current implementation for `radialGradient` && `linearGradient` contains many issues. Let me show it one by one.

1. `<radialGradient>` not works for `SVGLayeredImageView` at all. Always show empty result. (Reason: `SVGGradientLayer` override `drawInContext:` method will not works for CA render server, which is used for `SVGLayeredImageView` hardware-rendering.)
2. `<linearGradient>` in `SVGFastImageView` contains strange bounding rects, like the demo SVG images `firefox_logo_2013` below. (Reason: Apple's bug for `CAGradientLayer` on iOS. Works well on macOS, suck)
3. `<radialGradient>` does not supports standard [fx](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fx), [fy](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fy), [fr](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fr) properties,  cause many valid SVG render failed.

# Changes detail
+ Seperate the two class and follow SVG spec about the DOM interface [SVGRadialGradientElement](https://developer.mozilla.org/en-US/docs/Web/API/SVGRadialGradientElement) && [SVGLinearGradientElement](https://developer.mozilla.org/en-US/docs/Web/API/SVGLinearGradientElement).
+ Refactory the `SVGGradientLayer` implementations, to supports **correct** software-rendering for radial gradienrt && linear gradient. Thansk for [IJSVG](https://github.com/IconJar/IJSVG) this third-party framework for code reference. I reference his implementation and port it to SVGKit.
+ Supports `fx`, `fy`, `fr` on `<radialGradient>`

# Solved issue
1. Solved. because `SVGLayeredImageView` using CA render server and hardware rendering. It does not call any of your `renderInContext:` code. Apple's `CAGradientLayer` contains a type [kCAGradientLayerRadial](https://developer.apple.com/documentation/quartzcore/kcagradientlayerradial?language=objc) for radial gradient, So I use this for hardaware-rendering. However, it does not supports any focal point (suck Apple...). Anyway, at least all non-focal point radial gradient can now been rendering on  `SVGLayeredImageView`.
2. Solved perfectly.
3. Solved perfectly using `SVGFastImageView`. However, these properties will be totally ignored for `SVGLayeredImageView` (See the reason above)

# Compare between fix

See comments below since there are many images.

# Demo

I also update the iOS demo because it hard-coded to use `SVGFastImageView` and I want to test `SVGLayeredImageView` as well. However, to simplify the review from @adamgit, I create another branch for that changes (only about demo changes, not about Core Framework).

See changes in my fork branch : [bugfix_radial_gradient_demo](https://github.com/dreampiggy/SVGKit/tree/bugfix_radial_gradient_demo). After all review done, I can create MR again.


dreampiggy:
> Because for issue 1, we always get empty images before the fix using `SVGLayeredImageView`, so I just skip the compare result.

# For issue 2

Test file: [NewTux.svg](https://upload.wikimedia.org/wikipedia/commons/b/b0/NewTux.svg)

+ Before fix - `SVGFastImageView`:


+ Before fix - `SVGLayeredImageView`:
Empty :)

+ After fix - `SVGFastImageView`:


+ After fix - `SVGLayeredImageView`:

dreampiggy:
# For issue 3

Test file: [Mozilla_Firefox_logo_2013.svg](https://upload.wikimedia.org/wikipedia/commons/7/76/Mozilla_Firefox_logo_2013.svg)

+ Before fix - `SVGFastImageView`:


+ Before fix - `SVGLayeredImageView`:
Empty :)

+ After fix - `SVGFastImageView`:


+ After fix - `SVGLayeredImageView`:

Source/DOM classes/SVG-DOM/SVGHelperUtilities.m
```diff
@@ -502,8 +502,6 @@ +(CALayer *) newCALayerForPathBasedSVGElement:(SVGElement<SVGTransformable>*) sv
 		maskLayer.fillColor = [UIColor blackColor].CGColor;
 		maskLayer.strokeColor = nil;
 		gradientLayer.mask = maskLayer;
-		if ( [gradientLayer.type isEqualToString:kExt_CAGradientLayerRadial])
```
adamgit, line 505: What's this change doing?

dreampiggy: See changes in `SVGLinearGradientElement`, which use [kCAGradientLayerAxial](https://developer.apple.com/documentation/quartzcore/kcagradientlayeraxial?language=objc). And changes in `SVGGradientElement`, which use [kCAGradientLayerRadial](https://developer.apple.com/documentation/quartzcore/kcagradientlayerradial?language=objc)

This value previously is a user-defined value, but it break the support for `SVGLayeredImageView` rendering `<radialGradient>` element. Because QuarzCore does not recognize it as a valid type in [CAGradientLayerType](https://developer.apple.com/documentation/quartzcore/cagradientlayertype?language=objc)

adamgit:
Oh, I hate this part :). It has cost us so much time and debugging effort over the years, first uncovering Apple's bugs, then trying to workaround them ... without destroying performance.

Couple of thoughts:

 1. The RainbowWing svg in the demos/tests has historically been our go-to measure for "are gradients working?". Please verify your changes leave this rendering correctly (I believe it will, but it has historically been a great simple test for gradient and transparency issues, and the two normally come together to be a nightmare!)
 2. I remember writing the *Layered* and *Fast* imageviews partly to make gradients work for at least SOME people (although with the caveat that you lost access to some core rendering features). From the code changes, and your comments, I am not sure what the new status is of these two classes - can you summarise? (and do we need to refactor/redesign them?)
 3. ... in particular, it sounded as though the contract of *Layered* was being changed to no longer be purely CALayer hardware rendering? I might have misunderstood, but if that is happening, I'd want to add some extra flags / options / log messages to make it clear, because that's a HUGE difference for a lot of real-life scenarios.
 4. One of my go-to tests for Layered vs Fast is to pull out Wikipedia's INSANELY large world-map SVG (off the top of my head the full size thing takes up > 200MB of RAM to render!) and watch what happens. **off the top of my head!**: Fast view has a huge time-to-first-render because it gets held up inside Apple's inefficient CALayer internals - and uses loads of RAM - but is very fast afterwards. Layered is just as slow to start, but remains slow.

adamgit:
> # For issue 3
> * Before fix - `SVGFastImageView`:
> 
> 
> 
> * After fix - `SVGFastImageView`:
> 
> 
> 

Did you attach these images the wrong way around?

The "Fix" appears to have broken the gradient rendering! :)

dreampiggy:
# Some other demo from MDN SVG specs.

You can clone the `bugfix_radial_gradient_demo` branch. Then run it in both `SVGFastImageView` && `SVGLayeredImageView` by long-pressing in the collection view cell.

---
MDN's demo SVG about `<radialGradient>`: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/radialGradient#Result

# The demo on iOS

`SVGFastImageView`:



`SVGLayeredImageView`:





# The demo on macOS

dreampiggy:
@adamgit Apple's QuartZ (Core Animation, another name) contains two type of rendering engine. The software render and hardware render.

When you writing custom code by subclass `CALayer` and overridw [renderInContext:](https://developer.apple.com/documentation/quartzcore/calayer/1410909-renderincontext), then draw the layer on a `CGContext` (whatever the context from, such as [drawRect:](https://developer.apple.com/documentation/uikit/uiview/1622529-drawrect) method in UIView), your custom code will be called and render it on CPU. Any method call like `CGContextFillRect` will directlly trigger a draw on bitmap context hosted by backing store. After all draw code finished, the bitmap will be directly send to GPU as a OpenGL or Metal texture (bitmap).

When you don't subclass `CALayer` and using the built-in types of layers to composite (which means, composite large layer tree using many sublayers), and don't call any `drawRect:` method, just directlly set the layer on view and let it go. You are now using hardware rendering mode. All the layer information will be submitted to `Core Animation Render Server`, and Apple translate layer information into OpenGL or metal code running directly on GPU, not texture is needed.

So, what actually `SVGFastImagView` work is based on software-render mode. And what `SVGLayeredImageView` work is based on hardware-render mode. Both of two type is possible. (However, hardware-render does not allows you to write GSGL code or Metal code, means you're limited on customization, you can only composite)

You can check WWDC 2014 this pdf about how Core Animation works : https://devstreaming-cdn.apple.com/videos/wwdc/2014/419xxli6f60a6bs/419/419_advanced_graphics_and_animation_performance.pdf?dl=1

dreampiggy:
> > # For issue 3
> > 
> > * Before fix - `SVGFastImageView`:
> > 
> > 
> > 
> > * After fix - `SVGFastImageView`:
> > 
> > 
> 
> Did you attach these images the wrong way around?
> 
> The "Fix" appears to have broken the gradient rendering! :)

What you means `broken` ? Click the link and open Safari / Chrome and check the correct render result. This SVG contains `fx`, `fy`, `fr`, which previously does not support at all !

adamgit:
> @adamgit Apple's QuartZ (Core Animation, another name) contains two type of rendering engine. The software render and hardware render.
> 

Yes, I understand that, but I wasn't sure what impact these changes have on what combination of backends we're using, and where?

In particular ... should we perhaps rename SVGK[Layered | Fast]ImageView to SVGKImageView[ Hardware | Software ] ?

(I think that would be a lot clearer? Currently they're named the way they are because I thought people would most care about "does this export CALayers or not?" -- the original SVGKit implementation always used CALayers, and couldn't run fast, so for backwards compatibility it mattered. But now, years later, maybe it's time to be more explicit about what the purpose of the classes is: hardware vs softawre rendering?)

adamgit:
Your "before fix" image looks many times more correct than your "after fix" image does.

E.g. look at the snout. All the detail has been removed!

e.g. look at the splkey tail "feathers" on the bottom and right - most of the gradients have vanished

dreampiggy:
@adamgit I fix a stupid issue on transform via 0a9c8c9. See the image again. It's now match the SVG spec. However, the shadow issue seems related to `CALayer.renderInContext:`. But I've no idea how to fix it.

> Important

> The OS X v10.5 implementation of this method does not support the entire Core Animation composition model. QCCompositionLayer, CAOpenGLLayer, and QTMovieLayer layers are not rendered. Additionally, layers that use 3D transforms are not rendered, nor are layers that specify backgroundFilters, filters, compositingFilter, or a mask values. Future versions of macOS may add support for rendering these layers and properties.

After fix - `SVGFastImageView`:



What the standard render result (PNG format):



I think this one is more matching the standard render result, than the previous one.

dreampiggy:
I think we need to test and provide more images. Actually I think if we follow what SVG spec says, we'll get the nearst result of modern browser like Safari or Chrome. However, WebKit contains some internal hacks which is not SVG spec. And they can render some specify SVG using new feature in CSS3, which is beyond my ability to implement at all 😅

adamgit:
Note that we also have the entire official set of test. Images built in to 
the iOS demo, and ordered by feature - hundreds of them! - but they are 
great for testing individual features, and debugging.

For real life scenarios, the complex multi feature images like firefox. 
Logo, wikipedia world. Map, etc. Are great.

dreampiggy:
@adamgit Good point. I'll run the W3C test case locally to check the result. Maybe tomorrow (This is midnight in my local timezone... 😴 )

adamgit:
> I think this one is more matching the standard render result, than the previous one.

Looks better now, but it's still losing some critical details that were working originally. We mustn't remove features / reduce render quality when fixing/adding new features :).

In particular, the white gradient on lower half of the fox's nose appears to have been lost.

dreampiggy:
@adamgit Fix one more not obvious behavior for SVG spec: Any `cx`, `cy`, `r`, `fx`, `fy`, `fr` value which is < 1.0f, will be treated as percent value (WTF)...However, for `<path>` or `<rect>` element, they should not be treated as percent value. I fix it and it now looks perfect.

+ After fix - `SVGFastImageView`



---

For example, these two SVG files (from [MDN]( https://developer.mozilla.org/en-US/docs/Web/SVG/Element/radialGradient#Result)), they should render as the same result.

```xml
<svg width="120" height="120" viewBox="0 0 240 120"
   xmlns="http://www.w3.org/2000/svg">
  <defs> 
    <radialGradient id="exampleGradient"> 
      <stop offset="10%" stop-color="gold"/> 
      <stop offset="95%" stop-color="green"/> 
    </radialGradient>
    <radialGradient cx="50%" cy="50%" r="50%" fx="25%" fy="75%" fr="10%" id="flameGradient">
      <stop offset="0%" stop-color="white"/>
      <stop offset="10%" stop-color="yellow"/>
      <stop offset="95%" stop-color="red"/>
    </radialGradient>
  </defs> 
  <circle fill="url(#exampleGradient)" cx="60" cy="60" r="50"/>  
  <circle fill="url(#flameGradient)" cx="180" cy="60" r="50"/>
</svg>
```

```xml
<svg width="120" height="120" viewBox="0 0 240 120"
   xmlns="http://www.w3.org/2000/svg">
  <defs> 
    <radialGradient id="exampleGradient"> 
      <stop offset="10%" stop-color="gold"/> 
      <stop offset="95%" stop-color="green"/> 
    </radialGradient>
    <radialGradient cx="0.5" cy="0.5" r="0.5" fx="0.25" fy="0.75" fr="0.1" id="flameGradient">
      <stop offset="0%" stop-color="white"/>
      <stop offset="10%" stop-color="yellow"/>
      <stop offset="95%" stop-color="red"/>
    </radialGradient>
  </defs> 
  <circle fill="url(#exampleGradient)" cx="60" cy="60" r="50"/>  
  <circle fill="url(#flameGradient)" cx="180" cy="60" r="50"/>
</svg>
```

dreampiggy:
Now the newest firefox logo can also be correctly rendered as well. I'm running the all W3C test cases for `pservers-grad-*.svg`. My goal is that `Don't be worse than the current implementation`. At least there are no such strange bounding rect borders, which is unacceptable from our users(developers). (I don't know why current implementation cause this borders, but it does not appear on macOS in my macOS support branch...Apple's suck ?)

Test files: [Mozilla_Firefox_logo_2013.svg](https://upload.wikimedia.org/wikipedia/commons/6/67/Firefox_Logo%2C_2017.svg)

+ Before fix - `SVGFastImageView`



+ After fix - `SVGFastImageView`

adamgit:
Great!

The rect borders appear to be that you have debug mode enabled? That's a feature of the demo app that puts a coloured rect around every layer so you can see the bounds (especially useful when testing e.g. animations)

SVGKitLibrary/Demo-iOS/VCAllSpecImages.m
```diff
@@ -26,6 +27,10 @@ -(void)viewDidLoad
 		
 		self.svgFileNames = [NSMutableArray arrayWithArray: xcodeVirtualFolderSVGContents];
 	}
+
```
adamgit: Quick check: is this new feature documented somewhere in the demo app?

It's important enough (makes a huge difference for what people are seeing) that I'd want it to be very obvious to users that you can do it.

Maybe instead we should have a segmented bar at the top or bottom of the page that lets you switch between the two modes at app-level, isntead of having to keep doing the prss-hold thing?

adamgit: (also because the new way you've implemented this, seems like it would be easier/better for the user interface to be switching this back and forth glboally)

dreampiggy:
@adamgit You don't catup up my issue...Please try to clone the current 2.x branch, run the demo by yoruself to see the result. Or you can directlly create another new Xcode project, set up simple code which link `SVGKit.framework` and render a local SVG like something listed above.

It's not a DEBUG border or something can be controlled by defining a macro. The CALayer render result contains random borders for radial gradient && lienar gradient (However, does not appear on macOS in same code). It's a bug...So I think the issue 1 is really serious because it's not ready for production at all...

adamgit:
I'm not denying it's a bug :) - I was only pointing out a possible source of the rects, which might make it easier to figure out how/why this is happening (e.g. find the code that enables the debug mode, comment it all out, and see if the bug goes away?)

I don't have a mac with me today so can't build myself right now :(.

SVGKitLibrary/Demo-iOS/VCAllSpecImages.m
dreampiggy: @adamgit
> is this new feature documented somewhere in the demo app?

You mean long-press the collection cell and popup two menus `layer` or `fast` this feature ? It's added by this changes to see both two type of rendering result. Previouslly we only render all SVGs in the demo using `SVGFastImageView` (expect 3 hard-coded SVG here: https://github.com/SVGKit/SVGKit/blob/2.x/SVGKitLibrary/Demo-iOS/DetailViewController.m#L408). However, it's not so convenient to check (Need to edit and re-compile).

I think (or in my opinion), since we have two `SVGFastImageView` && `SVGLayeredImageview` representing two type of rendering engine, we'd better provide a way to check both of two. Adding a menu in the UI is the most easy way to do so...

If you means there are no UI tips or something to let user know we can long-press on the collection cell, I can add one tip, it's easy. 😅

dreampiggy:
@adamgit I previouslly think the same as you (I guess the strange boder is from some DEBUG sublayer or something subpath). However, after 2 hours debugging I found I'm wrong. It's a render issue because of Apple's `CAGradientLayer`. I create a new empty Xcode project, write some CGContext draw code and a simple `CAGradientLayer`, mix them together, then the strange border appear...I think it's time to fire a bug-report to Apple.

Interesting that the same code on macOS works well without any issue. Maybe this software-render code for linearGradient can be replaced with hardware-render in macOS (radialGradient can not, because Apple's radial gradient does not supports focal point...Sadly).

SVGKitLibrary/Demo-iOS/VCAllSpecImages.m
adamgit: i.e. stick one of these at the top of the screen:

https://i.imgur.com/MSovEk4.png

...but with the options "FastImageView" and "LayeredImageView"

NB: I like your change that sets the titlebar to say which you're using (this is really good for when people report bugs and say "I'm using Fast" but it turns out they were using Layered, and got confused - the screenshot will show us what they were actually using ;))

ALSO, EVEN BETTER: have the same thing when viewing the image itself, so that when debugging we can quickly flip back and forth between the two while looking at an image :).

dreampiggy: @adamgit The choice whether to use layer or fast is per-SVG level...And it's memorized until you long-press again...I can add a button, but however, there are already `Animate`, `Debug` two buttons on the navigation bar. It seems have no extra spaces...

Another issue is that current code about that detail pages does not support refresh or reload. If I want to support live-change from `SVGFastImageView` or `SVGLayeredImageView` I need to change too much current code (These code seems written long year ago 😅). So I think add a option in the previous page (The list page) is better...

dreampiggy: The choice whether to use `layer` or `fast` is per-SVG level...And it's memorized until you long-press again. I can add a button in the detail page, but however, there are already `Animate`, `Debug` two buttons on the navigation bar. It seems have no extra spaces...

Another issue is in the current implementation, the SVG detail pages does not support refresh or reload. If I want to support live-change from `SVGFastImageView` or `SVGLayeredImageView` I need to change some more current code (the code seems was written long year ago 😅). So I finally adding a option in the previous page (The list page)...

dreampiggy: Anyway, if you don't mind I change some more demo code, I can make the switch from `SVGFastImageView` && `SVGLayeredImageView` inside the detail page. Maybe it's time to remove the ancient code and do a little refactory.

Maybe another way to show the UI. The navigation bar will not show the title (to save us some spaces), but with a new navigation bar button.
If the button is showing `Layer`, it's now rendered as `SVGLayeredImageView`, click it the button then the current page will reloaded with `SVGFastImageView`.
If the button is showing `Fast`, it's now rendered as `SVGFastImageView`, click the button then the current page will reloaded with `SVGLayeredImageView`.

dreampiggy:
What about this way to toggle switch between `FastImageView` or `LayeredImageView` ? I think this one is better because you can load online SVG from our Demo (enter the URL at home page text filed), this one allows simple way to view two type of rendering result.

See this step (GIF):

adamgit:
Looks good to me

dreampiggy:
@adamgit 😅 Thanks for merging.

Actually, there are still little W3C gradient test cases not be passed (They are broken in previous version as well). I guess it's some edge case in my implementation to follow the SVG spec. I'll spend some time to fix it. Once I've done, I'll create a merge request and mention which one be fixed.

Thanks again !

PR #586, opened 2018-10-21
Fix the some gradient rendering edge case in W3C

dreampiggy:
This PR contains some edge-cases fix from W3C test cases. And most of cases in `pservers-grad-*` can be passed now.

There are still little cases which need much more work to do, both of them are also fail in the previous version, I can list here:

+ `<pattern>` is totally un-implementated, so `pservers-grad-03-b`, `pservers-grad-07-b` fail
+ ~~`<text>` element with gradient un-implemented (Maybe this one can be done), so `pservers-grad-08-b` fail, partial of `pservers-grad-11-b` fail~~ Done. Supports gradient for fill. But gradient for stroke is far more harder to implement (need create custom `CALayer`)
+ `spreadMethod` for gradient is un-implemented, all using `pad`. (This one is much harder to implement because Apple does not have API), so `pservers-grad-10-b`, `pservers-grad-14-b` fail
+ `color` property which is `currentColor` or `inherit` is un-implemented, so `pservers-grad-18-b` fail
+ `xlink:href` property not works for `fill` or `stroke` propety, so `pservers-grad-13-b`fail

Maybe I'll find some easy-to-fix test cases and try to fix it as well. If you want to merge it, please comment here and I'll check if there are no more changes to do.


dreampiggy:
😅@adamgit During fixing the radial gradient test case `pservers-grad-08-b`, I found that we don't implement the gradient for `<text>` element, So I create a simple implemntation using `CALayer.mask`.

And then I found there are also some un-implemented thing related to `<text>` cause that test fail (custom font does not work). Currently the `font-family`, `font-style`, `font-weight`, `font-stretch` does not implement, So I fix them together...It looks like that **One bug fixed, there will be 10 more bugs related to it**

Source/DOM classes/Unported or Partial DOM/SVGTextElement.m
```diff
@@ -144,22 +173,192 @@ - (CALayer *) newLayer
         label.anchorPoint = CGPointZero; // WARNING: SVG applies transforms around the top-left as origin, whereas Apple defaults to center as origin, so we tell Apple to work "like SVG" here.
     
 	label.affineTransform = textTransformAbsoluteWithLocalPositionOffset;
-	label.fontSize = effectiveFontSize;
-    label.string = effectiveText;
+    label.string = [attributedString copy];
```
dreampiggy, line 176: `CATextLayer` is just like a `UITextView`. No detailed cusotmization. Using `NSAttributedString` is much better than String because it supports stroke via `NSStrokeWidthAttributeName` && `NSStrokeColorAttributeName`

For now, this stroke can only supports color, but not gradient layer mask. I think maybe someone can rewrite this to using Core Text API to draw each glyph character on the screen and mix gradient. However, this need far more code and beyond my experience.

adamgit:
Merge conflicts .... !

dreampiggy:
@adamgit Solved.

It seems you don't have mac to see the render result. Maybe I should post the different between these fix for W3C test cases ? Please wait a minute.

---

Done. The correct render result, can use GitHub with your broswer (Chrome is recommaned, Web Standard :)) to check (Or check the bitmap result): https://github.com/SVGKit/SVGKit/tree/2.x/Demo-Samples/W3C_SVG_11_TestSuite/svg

Source/DOM classes/Unported or Partial DOM/SVGTextElement.m
```diff
@@ -78,11 +90,31 @@ - (CALayer *) newLayer
...
+        [attributedString addAttribute:NSForegroundColorAttributeName
+                                 value:(__bridge id)fillColor
+                                 range:stringRange];
+    }
+    if (strokeWidth != 0 && strokeColor) {
+        [attributedString addAttribute:NSStrokeColorAttributeName
+                                 value:(__bridge id)strokeColor
+                                 range:stringRange];
+        // If both fill && stroke, pass negative value; only fill, pass positive value
+        // A typical value for outlined text is 3.0. Actually this is not so accurate, but until we directly draw the text glyph using Core Text, we cat not control the detailed stroke width follow SVG spec
```
dreampiggy: `cat` => `can`, typo

dreampiggy:
+ `pservers-grad-24-f.svg`

> Left (iPhone 8) is this PR's render result, right (iPhone 7) is the current latest 2.x branch (after macOS support was merged).

dreampiggy:
+ `pservers-grad-17-b.svg` && `pservers-grad-20-b.svg` (They looks the same)

> The right (current 2.x branch) crash because of assert failed (gradient id does not exist, SVG spec says you should use fallback color)

dreampiggy:
+ `pservers-grad-08-b.svg`

> Sadly I can only implement the gradient **fill** for text, but not **stroke**. Which it's not simple to do using layer composition. You have to use custom draw of Core Text.
> Why the interesting left font render is because, that W3C test case use a custom font `Blocky`, which is `woff` and `SVG` vector font format. But however, iOS only support `ttf` font. So I grab one similar [from the Web](https://fonts2u.com/blocky-regular.font), to ensure `font-familiy` feature works. If you remove that font, just looks the same like right one, but with the correct gradient fill color

dreampiggy:
+ `pservers-grad-11-b.svg`

dreampiggy:
+ `text-font-02-t.svg`, `text-font-202-t.svg`, `text-font-203-t.svg`, `text-font-204-t.svg`

> Since previous we does not implement fallback of `font-family`, and does not implement these `font-style`, `font-weight`, `font-stretch` value. Many font specified W3C test failed. Here I just list some of them, they share the same result.

dreampiggy:
+ `text-font-204-t.svg`

> Fix the fallback color

Issue #592, opened 2018-10-26
iOS projects not compiling after last commit


dvanwinkle:
### New Issue Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/SVGKit/SVGKit/blob/2.x/CONTRIBUTING.md)
* [x] I have reproduced the problem while running in Xcode, and read the Console log, and acted upon any relevant error or warning messages
* [x] I have reproduced the problem while running in Xcode's Debugger, and used breakpoints to examine the actual values of variables vs the expected values
* [x] I am using CocoaPods, but have verified my podspec points to SVGKit v2.x

### Issue Description and Steps

iOS projects are no longer building after the recent commit to 2.x. It's trying to build SVGKExporterNSImage and cannot find the correct headers, as expected.

onnoweb:
PR: https://github.com/SVGKit/SVGKit/pull/593

PR #593, opened 2018-10-26
Correcting typos in exclude files.

onnoweb:


dvanwinkle:
Perfect! I will point to that commit really quick and see if it fixes it, but, I bet it will.

Nice catch!

dvanwinkle (approved):
Fixed issues

dvanwinkle:
Just tested locally, this fixed the issues found in #592

PR #597, opened 2018-10-29
Upgrade the framework version to 2.1.0

dreampiggy:
Upgrade the framework version to 2.1.0. After this PR was merged, we can create a git tag `2.1.0` and release the new version.

And I update the readme to introduce the macOS support for user.

### DIFF
--- a/Demo-iOS.xcodeproj/project.pbxproj
+++ b/Demo-iOS.xcodeproj/project.pbxproj
@@ -8,11 +8,11 @@
 
 /* Begin PBXBuildFile section */
 		3242C36E217CF0CD0013765F /* Blocky.ttf in Resources */ = {isa = PBXBuildFile; fileRef = 3242C368217CF0CD0013765F /* Blocky.ttf */; };
+		32AF7BBE217389FD00142396 /* SVGKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 529DADF11C415FF00069B076 /* SVGKit.framework */; };
+		32AF7BC021738A0300142396 /* SVGKit.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 529DADF11C415FF00069B076 /* SVGKit.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		32D3D5EC2176E3FA00ED20D8 /* Mozilla_Firefox_logo_2013.svg in Resources */ = {isa = PBXBuildFile; fileRef = 32D3D5E62176E3FA00ED20D8 /* Mozilla_Firefox_logo_2013.svg */; };
 		32D3D5F12176FA0700ED20D8 /* Mozilla_Firefox_logo_2013.png in Resources */ = {isa = PBXBuildFile; fileRef = 32D3D5F02176FA0700ED20D8 /* Mozilla_Firefox_logo_2013.png */; };
 		32D42245217AED3900CFD064 /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 32D42244217AED3900CFD064 /* LaunchScreen.storyboard */; };
-		32AF7BBE217389FD00142396 /* SVGKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 529DADF11C415FF00069B076 /* SVGKit.framework */; };
-		32AF7BC021738A0300142396 /* SVGKit.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 529DADF11C415FF00069B076 /* SVGKit.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		4FD8223F1AD057F000E419D3 /* arcs01.svg in Resources */ = {isa = PBXBuildFile; fileRef = 4FD8223D1AD057F000E419D3 /* arcs01.svg */; };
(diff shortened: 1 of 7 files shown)


adamgit:
I was expecting to also see pbxproj updates in this PR - otherwise the demo project will fail to build on new instdalls (Because its trying to reference the old libsvgkit-2.0.0 file)?

dreampiggy:
@adamgit This is the prepare before we release v2.1.0. Please have a check.

dreampiggy:
It seems the Demo project using the `SVGKit.framework`. That static library is a standalone build configuration and not related to Demo project...But anyway, I can update it.

image

dreampiggy:
@adamgit Any reason we generate the static library with that version number ? Why not just `SVGKit-iOS.a` ? I've joined other open source organization. The static library like SDWebImage or CocoaLumberjack does not append this version number.

image

adamgit:
I am not sure, but ... I think it's a standard that Apple/Xcode used?

I agree that if we could get safely remove the version from the name then it would make the library easier for us to maintain!

Demo-iOS.xcodeproj/project.pbxproj
```diff
@@ -104,7 +104,6 @@
 		663FCFF616C9AF3C00CCBFB3 /* Sample Licenses from svg-android project.txt in Resources */ = {isa = PBXBuildFile; fileRef = 663FCFF516C9AF3C00CCBFB3 /* Sample Licenses from svg-android project.txt */; };
 		663FD00416CAB16B00CCBFB3 /* groups-and-layers-test.svg in Resources */ = {isa = PBXBuildFile; fileRef = 663FD00316CAB16B00CCBFB3 /* groups-and-layers-test.svg */; };
 		6649E0AD1617479200AFE92A /* QuartzCore.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 6649E0AC1617479200AFE92A /* QuartzCore.framework */; };
-		6649E10D1617577000AFE92A /* libSVGKit-iOS.2.0.0.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 6649E10A1617574700AFE92A /* libSVGKit-iOS.2.0.0.a */; };
```
adamgit, line 107: Now there's no libSVGKit-iOS.x.y.z.a at all?

dreampiggy: That iOS Demo, using the `SVGKit.framework` for target link library. This `libSVGKit-iOS.2.0.0.a` actually does nothing than just a placeholder in the xcode project. 😅 

See this: 
image

dreampiggy:
> I agree that if we could get safely remove the version from the name then it would make the library easier for us to maintain!

This filed is safe. The number is just appended to file name...(This does not compiled to static library binary), but people who build static library by theirself, already know what the version is. So it's no need to add it at all (And cause we manual change that number each time release). I think we can just remove that. (actually this does not cause any issue....file name :))

See the static library from SDWebImage && CocoaLumberjack

image

image

dreampiggy:
@adamgit Hi. Good news. It seems that after I sending the email to claim the owner of SVGKit from CocoaPods forum. They add you and me to the CocoaPods trunk owner. 

```
➜ pod trunk info SVGKit

SVGKit
    - Versions:
      - 0.0.1 (2014-05-19 21:43:42 UTC)
      - 1.0 (2014-05-19 21:49:54 UTC)
      - 1.2.0 (2014-07-01 11:48:47 UTC)
    - Owners:
      - Mirko Olsiewicz <mirko.olsiewicz@olsiewicz.de>
      - adam <adam.m.s.martin+cocoapods@gmail.com>
      - Alexandros Salazar <asalazar@weather.com>
      - DreamPiggy <lizhuoli1126@126.com>
```

Now we can upload the podspec into the trunk. So that all CocoaPods user can use a release version instead of current branch dependency like `pod 'SVGKit', :branch => '2.x'`

So I'm waiting for your release of SVGKit 2.1.0. After you push a new git tag `2.1.0` (Don't append something like `v2.1.0`. because CocoaPods && Carthage use the git tag for version number). I can upload the Podspec to the CocoaPods for release.

dreampiggy:
@adamgit I upload the `2.0.0` version podspec to the CocoaPod Trunk. CocoaPods user can now use the 2.0.0 tag release in their podfile. https://cocoapods.org/pods/SVGKit

And, is there any pending PR to merge for 2.1.0 release ? Maybe we can start a git tag release. The 2.0.0 can works for Xcode 10 now, but it still seems far away from the current 2.x branch lastest commit.

adamgit:
Yes, I have everything setup to do a 2.1.0 tag + release - just waiting to
get the last few PR's merged + checked here :).

Version number / tag number I am planning to do the same as the 2.0.0
release - tag is "2.0.0" - ... NOT like the 1.2.0 release - where tag was
"v1.2.0".

NB: GitHub strongly encourages "v1.1.1" but I don't like having the "v" in
there, it has caused problems with other tools in the past. As I understand
it, you're saying tha both CP and Carthage agree here? :)

dreampiggy:
Both CocoaPods or Carthage, by default use the git tag for the raw version number.
If you tag a git tag use `v2.1.0`, people using the semversion will cause issue.

CocoaPods

```
pod 'SVGKit', '~> v2.1.0' # This will not work.
pod 'SVGKit', '~> 2.1.0' # This works
```

Carthage

```
github 'SVGKit/SVGKit', ~> 2.1.0
```

adamgit:
Yep, good. Released now.

PR #598, opened 2018-10-30
Don't throw exception out side the framework, using the error check before source usage

dreampiggy:
Don't try to throw exception to the out side of the framework. Current there maybe chance that during the parser, when a URL is illegal, a exception will throw out. And the exception will throw to user side and cause crash.

As a framework, we should handle this possible case because it's possible that the input SVG contains some invalid URL. We just assume the source is nil and let it show the blank image or correspond element.

And also the SVG spec say that `` src can not be loaded (Or our case, the `NSInputStream` can not be established, we should report error instead of runtime crash. See [Image loading errors](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Image_loading_errors).

This is a little patch, seems no API break for 3.x


adamgit (commented):
Before adding this we got a LOT of fake bugs where people had malformed / missing data sources and blamed the library for failing to show an SVG. People dont check logs enough :).

The exception forces devs to read the error, but also makes it easy for people to catch this in their code, and our demo code shows an easy way to present that when doing developer testing.

Off the top of my head.... Without an exception there is no way for users to see if an image is still waiting to load or has crashed? 

However... What if we remove the exception, return nil *and* updated svgk*imageview to display a "broken link" image when an image is nil? (this could be a simple vector that we store as a constant string inside the imageview classes as an implementation detail)?

dreampiggy:
@adamgit Emm...It's true that the Broswer show a placeholder for the `<image>` source which can not be found. Like this SVG in W3C:

https://github.com/SVGKit/SVGKit/blob/3.x/Demo-Samples/W3C_SVG_11_TestSuite/svg/conform-viewers-02-f.svg

Open in Safari && Chrome, view this SVG will show a placeholder image.

If people are interested in this SVG parse failed for URL. We can use `SVGKParser` to report this parser result, and using a delegate to report the issue. And maybe I agree with you that we can inline a placeholder string for this case in `<image>` element (For other element, just only use the log).

But it's not a good idea to throw a exception outside. Because that method seems a common method (Can also been called from public API). Exception is mostly the `unexpceted result and dangerous happend.` in Objective-C world (Not in Swift). We;d better using `NSError` for expceted error result to report.

dreampiggy:
@adamgit Please check again. I add the fallback broken image inline resource. It's from the Safari..Actually this icon:

icon

This PR also fix one issue that `<text>` element does not allow `x`, `y` to be percent value, which should allow as SVG spec. Fix the test case `conform-viewers-02-f.svg`

dreampiggy:
This looks the same as Safari WebKit now. No crash anymore.

iOS:



Safari:

adamgit:
OK, looks good. But ... Can you send this PR against 3.x instead please? I don't want to be merging to 2.x unless there are urgent/security/etc critical bugs - I would prefer all new dev is on 3.x branch.

dreampiggy:
😅 I think this is a patch version release, since no new feature was introduced, it's a patch to solve that the throw-out exception.

I think as a framework, we'd better follow semversion: https://semver.org/, only the API-breaking release should be marked as 3.x.

dreampiggy:
I can imagine one 3.x main changes, it that to solve the Core DOM symbols. Current that out-of-prefix DOM class should using the correct prefix to avoid conflict with the naming. (Objective-C's pain).

Current here are already conflict names:

+ `DOMImplentation`, which Apple's `WebKit.framework` contains this symbol, we named it `AppleSuckDOMImplementation` (I think the author don't really know Objective-C, the comments there point to something related to `App Store` ??? but it's not what the issue is)
+ `Comment`, which Apple's `CarbonCore.frameowork` contains this symbol, we named it `SVGKComment`.

So it's better, to rename it all correctly, with prefix + correct DOM (W3C standard) naming.

dreampiggy:
@adamgit Changed the base branch to 3.x

PR #599, opened 2018-11-01
All init method should call `populateFromImage` to correctly setup the layer class (layer-backed or layer-hosting)

dreampiggy:
This fix issue if user write this case (really common and no warning):

```objective-c
SVGKImageView *imageView = [SVGKImageView alloc] initWithFrame:frame];
```

Or using storyboard on macOS. Because the `layerClass` is not correct been setup (The setup logic is inside `populateFromImage:`). So it wll not use `SVGKLayer` but the AppKit default layer. And then cause a runtime crash.

### DIFF
--- a/Source/UIKit additions/SVGKFastImageView.m
+++ b/Source/UIKit additions/SVGKFastImageView.m
@@ -39,12 +39,7 @@
 	self = [super initWithFrame:frame];
 	if( self )
 	{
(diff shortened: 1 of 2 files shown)


dreampiggy:
@adamgit

adamgit:
Everything should be going to the 3.x branch now ... I've merged this, but presumably needed in 3.x too?

Issue #591, opened 2018-10-25
SVG file displaying error - "clipPath" element problem


lukaszkrystek:
### New Issue Checklist

* [+] I have read and understood the [CONTRIBUTING guide](https://github.com/SVGKit/SVGKit/blob/2.x/CONTRIBUTING.md)
* [+] I have reproduced the problem while running in Xcode, and read the Console log, and acted upon any relevant error or warning messages
* [+] I have reproduced the problem while running in Xcode's Debugger, and used breakpoints to examine the actual values of variables vs the expected values
* [+] I am not using CocoaPods, or: I am using CocoaPods, but have verified my podspec points to SVGKit v2.x

### Issue Description and Steps

Please fill in the detailed description of the issue (full output of any stack trace, compiler error, ...) and the steps to reproduce the issue.

Some SVG files don't display properly.
They display properly in browsers (Chrome, Safari). However, using SVGKit
(iOS/macOS) they do not display correctly.
Please see attached files.

[SVG_not_supported.zip](https://github.com/SVGKit/SVGKit/files/2515010/SVG_not_supported.zip)

adamgit:
Thanks. Can you be specific about which features aren't displaying?

The problem here is that if you don't say WHAT isn't rendering, no-one
will find it when searching for particular features / issues they want
to work on.

And also: no-one has any idea if it's an area they want to fix.

lukaszkrystek:
Ok, It's my error - only one attached file does not display - icon_invite.svg
Other looks okey in the newest version of SVGKit.

I've attached screenshots, how it looks like in browsers (Chrome, Safari) and how it looks like in Demo-iOS application.

[screenshots.zip](https://github.com/SVGKit/SVGKit/files/2515217/screenshots.zip)

Thanks in advance.

adamgit:
If you attach the files individually, it's more likely someone will
look into it.

If you spend some effort in looking at your SVG files and working out
which path/command/SVG tag is not rendering correvtly, it's about ten
times more likely someone might try to fix it.

lukaszkrystek:
OK thanks - I will try to find which tag does not render correctly. Also I attached screenshots individually.

<img width="1227" alt="chrome - ok" src="https://user-images.githubusercontent.com/13151687/47508095-bee02780-d873-11e8-979c-0d388c358e78.png">

[...]

PR #601, opened 2018-11-02
Use NSScanner to tokenize class names instead of -[NSString enumerate…

### LINKED ISSUE #600: SVGKit incorrectly handles hyphens (-) in selector names
### New Issue Checklist

* [x] I have reproduced the problem while running in Xcode's Debugger, and used breakpoints to examine the actual values of variables vs the expected values
* [x] I am not using CocoaPods, or: I am using CocoaPods, but have verified my podspec points to SVGKit v2.x

The CSS spec [allows hyphens in selector names](https://www.w3.org/TR/CSS21/syndata.html#characters). However, when SVGKit's SVGElement.m tries to decide whether a selector applies to an element, it uses -[NSString enumerateSubstringsInRange:options:usingBlock:] to tokenize the string. The problem is, with the NSStringEnumerationByWords option, this tokenizer considers Hyphenated-Words to be two words, not one. So if the CSS has a hyphen in the selector name, it'll fail to match every time.

Chrome and Safari handle selectors with hyphens in them correctly. I have a proposed fix that will correctly parse selectors with hyphens in them; stand by for a pull request.

nickzman:
…SubstringsInRange:options:usingBlock:], since that method tokenizes strings with hyphens in them into separate strings, causing the method to fail if there is at least one hyphen in the selector name. Add a sample that illustrates the problem. Without the change, the two half-circles in the bottom-right corner are filled with black (incorrect), whereas with the change, they're filled with bright red (correct).

Fixes #600


dreampiggy:
From the SVG spec, the `class` property is defined like this:
>  lists within SVG's XML attributes can be either comma-separated, with optional white space before or after the comma, or white space-separated.

Your changes will fix the `-` separator. But you should also filter the comma ',' charactor, which should seperate the substrings.

I agree using `enumerateSubstringsInRange` this method is wrong, because Apple will always use any charactor like `!@#$%^&*()`. It's better to use the SVG spec listed separator.

So, could you please change your code, to also filter the comma `,`, as well as the white space as separator ? (Using the `characterSetWithCharactersInString:` method and add all SVG spec charactor code like that Unicode whitespace && comma into them).

nickzman:
I did it using NSMutableCharacterSet just to be thorough in covering what counts as a space.

adamgit:
New commit looks good to me - @dreampiggy ? I believe the only reason its been left using enumerateSubstringsInRange for so long is that Apple's docs for that method are missing so no-one was 100% sure what it does, and it was risky to replace it :). I just checked the docs on apple.developer.com and ... still missing. Lets get rid of it!

If not, can you share the spec link you're using? Off the top of my head, this attribute is half defined in SVG and half defined in CSS. 

(for reference, here's the part from CSS2: https://www.w3.org/TR/CSS2/syndata.html#characters

> In CSS, identifiers (including element names, classes, and IDs in selectors) can contain only the characters [a-zA-Z0-9] and ISO 10646 characters U+00A0 and higher, plus the hyphen (-) and the underscore (_); they cannot start with a digit, two hyphens, or a hyphen followed by a digit. Identifiers can also contain escaped characters and any ISO 10646 character as a numeric code (see next item). For instance, the identifier "B&W?" may be written as "B\&W\?" or "B\26 W\3F".
)

dreampiggy:
Agree that `enumerateSubstringsInRange:` seems not really documented what it does. From the test, it will split the string using many characters inside ASCII which is not alphabet or numbers...We'd better use the accurate filter which follows SVG spec.

LGTM now.

dreampiggy:
@nickzman @adamgit  Can we change the target branch to 2.x (Current 3.x && 2.x are no difference on code. Just some documentation update). This is like a patch version to 2.x, but not 3.x new features.

dreampiggy:
@adamgit @nickzman This also fix the issue of #591. I suggest to merge this into 2.x.

And we can release a patch version such as 2.1.1 to solve these issue. They are absolutely bug, not feature or API breaking changes.

In 3.x we can add some new API break changes, I think it's not now to change anything base branch into 3.x

adamgit:
We don't generally have the resource to support two master branches :(.

Our mainline branch (could be called "master" but named 2.x, now 3.x) is generally very high quality, and everyone should be using it in production. I would rather push people to use 3.x. Most committers cannot/will not make two copies of every PR. 

Our only reason for keeping 2.1.0 alive is for people who cannot afford to follow the latest. My goal with 3.x is to immediately start merging any PRs that e.g. lose compatibilibty with massively outdated versions of xcode/iOS.

nickzman:
So what's the consensus? Do I need to rebase this on 2.x, or can one of you merge this into the 3.x branch, and maybe cherry-pick it back into 2.x if you want?

adamgit:
If you rebase on 3.x, I'll merge it directly.

I want to start pushing 3.x and 3.x PR's ASAP :).

dreampiggy:
@adamgit So, you mean this SVGKit does not follows the [semver](https://semver.org/) ?

Is there any new API-breaking changes was introduced and will merged into 3.x branch ? I thinik If there are no API-breaking changes, what about release a patch or minor version update. The version number is nothing but the semantic about API changes to users, I think.

dreampiggy:
@adamgit I think we can just keep one branch for master.

If you have API break (As a PR, we both can check this through public header modification). I think we can just merge it into 3.x (or another branch like we called `future` branch).

If you don't have any API break, just a patch, you can merge it into 2.x.

The 3.x branch just update as needed to follow the 2.x branch (using merge is OK), so if someday we found nothing else patch to 2.x, we merge the 2.x to 3.x last time, and then drop 2.x support, use 3.x for new feature. And at that time we may also create a 4.x branch for API-breaking merge request, etc.

This is what we've done in the previous of SDWebImage. It actually cost nothing if there are no any API-breaking changes was introduced (Because your 3.x branch is stable until someone put a API-breaking change).

Currently all pending issue seems no API-breaking changes, what branch is was merged is OK since we can always sync them. But if someone fire a API-break changes, if we don't have two branches, how can we solve the end-user's dependency ? (You mean using the `3.x` branch for production, however, if the API changed without any annotation, may cause the downstream software break)

adamgit:
This is not the place to have an extended discussion about branching strategy, or to change the established strategy that's been in place for almost 5 years.

I will look into deleting/removing the 2.x branch (the only reason I haven't already is because I wasn't sure if this would have unwanted side-effects). Any fixes to 2.x should be branched direclty from 2.1.0. There is no need for a 2.x master now.

All new and outstanding PRs - as per the README, as per the front page, and as per the GitHub current setup - should be issued against 3.x. That is the one, single, master development branch. API breaking/changing changes *will* be accepted if they have good reason.

dreampiggy:
So, this can answer my question: `SVGKit does not follows the semver` ?

If so, my SVG coder plugin [SDWebImageSVGCoder](https://github.com/SDWebImage/SDWebImageSVGCoder) , which depends on SVGKit, have to lock the version number of SVGKit for each build. Because we follows the semver, but the dependency does not.

nickzman:
Gentlemen, you can't fight in here! This is the War Room! ;)

Seriously, this PR is already targeting the base (3.x) branch, so all I need is for someone to merge it if there are no other code changes necessary. Then you can do whatever you want with the 2.x branch.

dreampiggy:
@nickzman  😅  It's not fight...I want to ensure whether this framework follows semversion or not. So that my PR can do some changes to match it.

Previouslly I assume SVGKit follows semversion, so that I'm carefully not to break public API during some PRs, though those APIs are wrong. If we does not follows semversion, we can do refactory and fix them at anytime.

Whether to merge 2.x/3.x is not different if we does not follows semversion. It's just a develop branch, and the number does not means anything.

dreampiggy:
So we enter an aggrement that SVGKit does not follows semversion ? I can also update my PRs to match the target branch to 3.x. And it's good to merge 2.x into 3.x and then drop it.

Some PRs may also change the public API at any time, if so, I suggest to make an annotation in the change log, or the user may find it hard to use a statble version of SVGKit.

PR #602, opened 2018-11-06
Use a better solution to solve macOS's flip of <image> && <text> contents

dreampiggy:
Don't need to recursive check and hack for that transform

## What's for

This solution provide a better way, instead of current hack for macOS. See that `temporaryWorkaroundPreprocessRenderingLayerTree:` method.

The reason, is that we use the iOS flipped coordinate system on macOS, instead of native bottom-left coordinate system (Because it can simplify cross-platform code). This works for most of rendering of vector element, because there are a global transform from AppKit. However, for the bitmap or texture element like `<image>` && `<text>`, They are kept into the default coordinate system (bottom-left), so the final render result will cause them been flipped. 

Note this only happened to `SVGKFastImageView` software rendering, while `SVGKLayeredImageView` already handle this case. Good Apple :)

## The solution

For text element rendering in `CATextLayer`, it can override that `contentsAreFlipped` property to use the flipped coordinate system.

For image element, there are no obvious way to do so, we still use the `contentsTransform` property. (Actually, the CALayer set a flag in CALayer->_ivars to indicate it's flipped , but it's hard to use that than `contentsTransform`, so I still use this one, same effect)

This solution is better than the current one, which do not need to recursive check each sublayers. Just render them on demand.

@adamgit


dreampiggy:
This solution changes, also fix the current macOS demo usage. Which provide a feature to let you view the different render result on two views at the same time.

However, the implementation of this feature (we share `SVGKImage` instance) will cause a complicated use case: When you firstlly rendering a SVG in `SVGKLayeredImageView`, then reusing the cached CALaye on another `SVGKFastImageView`, the image && text remind flipped (Because Core Animation set a flag to indicate it's flipped, we use another `contentsTransform` will re-flip it again... 😅 )

Actually I don't think this usage is correct. A CALayer should only associated to one View, but not reusing and rendered in different views at the same time. But anyway, this solution can solve the problem as well.

dreampiggy:
@adamgit Changed the base branch to 3.x

PR #607, opened 2019-01-08
Fix warnings of `GCC_WARN_ABOUT_MISSING_NEWLINE` for 2.x branch

NSHaoSong:


NSHaoSong:
@adamgit we turned on `GCC_WARN_ABOUT_MISSING_NEWLINE` in our project, so these headers are causing build warnings. Also porting the same changes here #608 for 3.x branch.
Could you let me know if these changes look OK? Thanks!

adamgit:
I don't see a need for this feature - if you have tools that require a newline  at the end of file, you should fix your tools! :) - but I believe that our header files should aim to be as easy to use as possible, so I'm accepting this.

PR #608, opened 2019-01-08
Fix warnings of `GCC_WARN_ABOUT_MISSING_NEWLINE` for 3.x branch

NSHaoSong:
Apply fixed from #607 to 3.x branch.


adamgit:
I don't see a need for this feature - if you have tools that require a newline  at the end of file, you should fix your tools! :) - but I believe that our header files should aim to be as easy to use as possible.

I'm happy to have the header files with extra newline, but please don't change the compiler warnings in the main project.

NSHaoSong:
Sure! I will revert the changes in 2nd commit.

PR #609, opened 2019-01-15
don't parse url if it is nil

klein-artur:
In some SVGs the creator can make the misteak to place an object with a href that points to a local path. In this case, the SVGKit parsing crashes, because the url will be nil.

It should check if the url provided by the href is valid (NSURL pointer is not nil) and therefore try to fetch that url, or just ignore it (the broken image will be shown anyway)

A said svg example is added. unfortunately as a txt, because svg is not supported. Just change the file extension to svg.

[svgimg.txt](https://github.com/SVGKit/SVGKit/files/2761245/svgimg.txt)

### DIFF
--- a/Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGImageElement.m
@@ -104,7 +104,7 @@ CGImageRef SVGImageCGImage(UIImage *img)
     {
         SVGKImage *svg = nil;
         
-        if( effectiveSource == nil )
+        if( effectiveSource == nil && imageURL != nil)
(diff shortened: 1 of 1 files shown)


adamgit:
Good idea, but shouldn't we instead (or in addition?) make "SVGKSourceURL sourceFromURL" return nil for a nil input URL?

If it returned nil (i.e. sourceFromURL is invalid if its nil) then this would automatically work already.

And ... it seems to me that an SVGKSource constructed with nil URL should not be valid/allowed anyway?

klein-artur:
Sorry, just saw that I used a 2.x branch, and that in the the 3.x branch the SVGKSourceURL sourceFromURL already returns nil if the url is nil.

PR #612, opened 2019-02-15
Feature: optional custom cache name on SVGKImage init

jfspencer:
I am wrapping up an integration with React Native and needed a way to customize the cache key name and reset the cache as needed.

Checked these changes against the Demo-iOS project and found everything working. This PR  increases init API surface on SVGKImage, ~and concats any customKey string with the svg name~ by adding additional init methods that will take a cache key as a parameter.

_added_
+(void) clearCache public method for  +(void) didReceiveMemoryWarningOrBackgroundNotification

+(SVGKImage *)imageNamed:(NSString *)name withCacheKey:(NSString *)key

_modified_
+(SVGKImage *)imageNamed:(NSString *)name inBundle:(NSBundle *)bundle withCacheKey:(NSString *)key
checks for `key` and uses that as the cache key, if none is provided then `name` is used as the cache key as before

didn't see a need to include an svg for this.

### DIFF
--- a/Source/SVGKImage.h
+++ b/Source/SVGKImage.h
@@ -88,7 +88,9 @@ typedef void (^SVGKImageAsynchronousLoadingDelegate)(SVGKImage* loadedImage, SVG
  - Creates an SVGKSource so that you can later inspect exactly where it found the file
  */
 + (SVGKImage *)imageNamed:(NSString *)name;
++ (SVGKImage *)imageNamed:(NSString *)name withCacheKey:(NSString *)key;
 + (SVGKImage *)imageNamed:(NSString *)name inBundle:(NSBundle *)bundle;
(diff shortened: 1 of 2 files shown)


adamgit:
didReceiveMemoryWarningOrBackgroundNotification shouild call clearcache, not the other way around.

cacheKey should replace name, not append to it (if you're letting people specify a cache-key, let them specify it!)

jfspencer:
@adamgit good points, thanks :) . made the adjustments.

jfspencer:
@adamgit anything else needed for this to get merged in?

adamgit:
Is there a reason to include:

[self didReceiveMemoryWarningOrBackgroundNotification: nil];

inside clearCache?

jfspencer:
@adamgit nope, my oversight, just removed it.

adamgit:
Github is broken. The merge button cannot be clicked

(I can click it, but nothing happens).

I've emailed github but they're not responding :(.

PR #613, opened 2019-02-19
# SVGElement position issue addressed.

santu1990:
- Considered origin x & y for **svg** element.

Notes : I would like add few points.

- I was wondering why we simply ignored  **x_Pos** & **y_Pos** for element. I figured it out we are silently ignoring position value. I created PR with changes, let me know your thoughts.
- I am using svg with multiple embedded **svg** .

### DIFF
--- a/Source/DOM classes/Unported or Partial DOM/SVGSVGElement.m
+++ b/Source/DOM classes/Unported or Partial DOM/SVGSVGElement.m
@@ -158,12 +158,12 @@
     NSString* pos_y = [self getAttribute:@"y"];
     
     if (pos_x == nil || pos_x.length < 1)
-        self.x = nil;
+        self.x = 0; // i.e. undefined
     else
         self.x = [SVGLength svgLengthFromNSString:pos_x];
     
     if (pos_y == nil || pos_y.length < 1)
-        self.y = nil;
+        self.y = 0; // i.e. undefined
     else
         self.y = [SVGLength svgLengthFromNSString:pos_y];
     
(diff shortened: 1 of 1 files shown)


adamgit:
If "self.x == nil || self.y == nil" - shouldn't we fall back to the old code that assumed 0 for either/both?

...or will that cause other problems?

I approve of your changes, but I'm concerned that this could break SVG's that were already working if they merely fail to specify x and/or y?

santu1990:
@adamgit : I see your point, I would have same point in before my PR, does it really make sense to process for element which has invalid position(`"self.x == nil || self.y == nil"`). Let me know your thought or do we have any SVG with this use-case, I can test it against fix.

**OR** we can assign **X & Y** to **0** as it was working earlier.
```
if (pos_x == nil || pos_x.length < 1)
        self.x = 0;
    else
        self.x = [SVGLength svgLengthFromNSString:pos_x];

    if (pos_y == nil || pos_y.length < 1)
        self.y = 0;
    else
        self.y = [SVGLength svgLengthFromNSString:pos_y];
```

adamgit:
"Does it make sense with invalid position" ? -- yes, because a missing position is not "invalid". In most of SVG (and HTML/CSS) a missing value is defined as being either "0", or "auto", or "inherit the parent element's value".

Originally, we assumed 0 for width and height also --- HOWEVER, there are a few edge-cases where the difference between "no rectangle for this element" and "a 0-width rectangle for this element" actually make a difference, so we had to create the nil check for width & height.

So I think your new suggestion - i.e. "we can assign X & Y to 0 as it was working earlier." - is better here.

adamgit:
e.g. here in the official SVG spec: https://www.w3.org/TR/SVG/geometry.html#XProperty

... x defaults to "0" :)

santu1990:
@adamgit : I updated the PR with changes, please have a look, let me know if further changes needed, we can merge if it is fine :)

PR #615, opened 2019-02-20
The iOS targets have assertions turned on for the Release configuration. I'm proposing that assertions be turned off for that configuration at the project level.

chriseplettsonos:
…d project file level disabling.

This fixes the iOS targets, which did not have asserts disabled for the Release configuration.


adamgit:
Github is failing to merge this -- appears to be a new bug in github :(.

Issue #641, opened 2019-08-22
Crash on [SVGKParser addParserExtension:] + 408


neobie:
I am using SDImage with SVGKit on UITableView.
Notice a huge crash logs from Fabric. Can anyone help?

```
SDImageSVGCoder *SVGCoder = [SDImageSVGCoder sharedCoder];
            [[SDImageCodersManager sharedManager] addCoder:SVGCoder];
            CGSize SVGImageSize = CGSizeMake(20, 15);
            [cell.imageFlag sd_setImageWithURL:[NSURL URLWithString:urlString] placeholderImage:nil options:0 context:@{SDWebImageContextSVGImageSize : @(SVGImageSize)}];
```

Crashed: com.hackemist.SDWebImageDownloaderOperationCoderQueue
0 libobjc.A.dylib 0x1e2f93bf0 objc_retain + 16
1 SVGKit 0x104e627ec -[SVGKParser addParserExtension:] + 408
2 SVGKit 0x104e625f0 -[SVGKParser addDefaultSVGParserExtensions] + 284
3 SVGKit 0x104e62354 +[SVGKParser newParserWithDefaultSVGKParserExtensions:] + 100
4 SVGKit 0x104e62380 +[SVGKParser parseSourceUsingDefaultSVGKParser:] + 24
5 SVGKit 0x104e5f6f0 -[SVGKImage initWithSource:] + 64
6 SVGKit 0x104e5f8b4 -[SVGKImage initWithData:] + 184
7 SDWebImageSVGCoder 0x104e2b880 -[SDImageSVGCoder decodedImageWithData:options:] + 100
8 SDWebImage 0x104d8755c -[SDImageCodersManager decodedImageWithData:options:] + 348
9 SDWebImage 0x104d89f3c SDImageLoaderDecodeImageData + 828
10 SDWebImage 0x104d940b4 __70-[SDWebImageDownloaderOperation URLSession:task:didCompleteWithError:]_block_invoke.198 + 216
11 libdispatch.dylib 0x1e37e0a38 _dispatch_call_block_and_release + 24
12 libdispatch.dylib 0x1e37e17d4 _dispatch_client_callout + 16
13 libdispatch.dylib 0x1e378a324 _dispatch_lane_serial_drain$VARIANT$mp + 592
14 libdispatch.dylib 0x1e378ae40 _dispatch_lane_invoke$VARIANT$mp + 428
[...]

PR #642, opened 2019-09-06
Fix the crash issue #551 by using the instance property instead of static variable.

dreampiggy:
Also fix the protocol method implementation, which should return NSArray instead of NSSet

CC @adamgit 

See the issue herer: #551 #641
See the reason that cause this crash: https://github.com/SVGKit/SVGKit/issues/551#issuecomment-527986599

### DIFF
--- a/Source/Parsers/Parser Extensions/SVGKParserGradient.m
+++ b/Source/Parsers/Parser Extensions/SVGKParserGradient.m
@@ -16,24 +16,25 @@
 #import "SVGLinearGradientElement.h"
 #import "SVGRadialGradientElement.h"
 
+@interface SVGKParserGradient ()
+@property (nonatomic) NSArray *supportedNamespaces;
+@property (nonatomic) NSArray *supportedTags;
+@end
+
 @implementation SVGKParserGradient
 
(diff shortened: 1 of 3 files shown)


Source/Parsers/Parser Extensions/SVGKParserGradient.m
```diff
@@ -16,24 +16,25 @@
...
 #import "SVGRadialGradientElement.h"
 
+@interface SVGKParserGradient ()
+@property (nonatomic) NSArray *supportedNamespaces;
+@property (nonatomic) NSArray *supportedTags;
+@end
+
 @implementation SVGKParserGradient
 
-//-(NSSet*) supportedNamespaces
```
dreampiggy, line 21: This required protocol method does not implements. And, all the following return value should be `NSArray`, but not `NSSet`. See the protocol definition.

adamgit:
Most of the code being changed here is more than 6 years old, and looks like it was committed only as a "temporary improvement" on the previous broken code. So although these changes might slightly break some 3rd party code, I think it's worth changing - the existing code wasn't robust anyway!

PR #648, opened 2019-10-10
removing cocoalumberjack dependecy from latest version

lagubull:
WE WON't MERGE this, it is only for visibility of the changes

### DIFF
--- a/SVGKit-iOS.xcodeproj/project.pbxproj
+++ b/SVGKit-iOS.xcodeproj/project.pbxproj
@@ -3329,8 +3329,7 @@
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
-					"$(PROJECT_DIR)/Carthage/Build/Mac",
-					"$(PROJECT_DIR)/3rd-party-frameworks/CocoaLumberjack-2.2.0/Mac",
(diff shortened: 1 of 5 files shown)


danieltmbr:
Why don't we merge? what's the purpose of the PR then?
Also why do we need to remove CocoaLumberjack?

lagubull:
I don't want to merge it so that we can keep 3.x free of any TW changes so we can simply fetch from upstream whenever we need to.

lagubull:
cocoaLumberJack is used in other pods and that causes issues according to XCode warnings...

adamgit:
I also don't understand the point of this, it would be great if you could
give a slightly more detailed description.

AFAICS:
 1. This unlinks Lumberjack
 2. This removes Lumberjack from Cocoapods
 3. The purpose is ... ?? to workaround a bug in Cocoapods ??

If Cocoapods is preventing you from managing LJ versions easily and in
*compatible, non-proprietary* ways, that sounds like a huge bug in CP. Is
that the case?

I know it happened in the past (it's one of the reasons we banned CP from
projects at multiple companies I worked at), but I thought they fixed those
bugs long ago? Surely CP has this covered by now?

lagubull:
Sorry, this PR was meant to go to my internal 3.x branch. not public

adamgit:
Ah! LOL OK, no worries :).

PR #653, opened 2019-10-17
fix error of M-to-A.svg

Penny1029:


adamgit:
Should this change be made to most of the other commands too?

If I understand correctly, the issue is that an M command can jump straight into an A ... but if so, why doesn't this break other commands (I'm sure we've had many examples of this M-to-A before in live SVGs). Is it a bug in the elliptical curve code specifically?

PR #656, opened 2019-11-13
2.x make SVGKit-iOS target safe for extensions

luisafcastano:
I previously made the `SVGKitFramework-iOS` target safe for extensions, but realized today that what Carthage uses is the SVGKit-iOS target. Apologies for that mistake.


adamgit:
This is a merge against the old 2.x branch. Apart from emergency backports, we're not modifying that branch - all commits need to be against 3.x branch.

luisafcastano:
I understand. In our `Cartfile` we generally only point to stable versions (`2.1.0` for SVGKit) or to commits we do to the latest stable version, that's why my changes were done against `2.x`.

To extend the change to the upcoming version, I have created a [PR](https://github.com/SVGKit/SVGKit/pull/657) that applies the safe-extension-safety changes to `3.x`.

Feel free to reject this PR and only look at the other one. I will then close this one.

Thank you!

adamgit:
We should probably push a 3.0 release then too :)

PR #658, opened 2019-11-22
add_svgelement_transform

lichentao:
Description:
      <g> element and sub element  has implement transform,  svg   can Apply transform ? for test ok 

Example:
<svg viewBox="0 0 352.2 372.8" transform="scale(2.0)"  xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">

### DIFF
--- a/Source/DOM classes/SVG-DOM/SVGDefsElement.m
+++ b/Source/DOM classes/SVG-DOM/SVGDefsElement.m
@@ -1,5 +1,4 @@
 #import "SVGDefsElement.h"
-
 @implementation SVGDefsElement
 
 @end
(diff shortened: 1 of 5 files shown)


adamgit:
The following elements should have "SVGTransformable":

SVG, G, A, DEFS, USE, IMAGE, SWITCH, PATH, RECT, CIRCLE, ELLIPSE, LINE, POLYLINE, POLYGON, TEXT

It appears to be missing from:

SVG, DEFS, SWITCH

Please could you add this change to the SVGDefsElement.* and SVGSwitchEleemnt.* as well, so that all places are now correct?

Also, please include the one-line comment in the .m files where you have the line "@synthesize transform;" -- all the other classes have the comment:

""synthesize transform; // each SVGElement subclass that conforms to protocol "SVGTransformable" has to re-synthesize this to work around bugs in Apple's Objective-C 2.0 design that don't allow @properties to be extended by categories / protocols"

...this will make sure no-one accidentally deletes that @synthesize line (unless it is no longer necessary? I haven't checked recently)

PR #666, opened 2020-01-17
Element add cssstyle

lichentao:
介绍： 通过给style注入css字符串与Element映射
使用方法：1.注入CssStyle样式中添加样式id 或者 class
                    2.给element vdom 标签设置 id/class


adamgit (changes requested):
1. SVGSVGElement does *not* extend SVGTransformable -- https://www.w3.org/TR/SVG11/struct.html#InterfaceSVGSVGElement -- that would break the spec

2. styleHasRule is being continuously overwritten. Perhaps this line should be "styleHasRule = styleHasRule | [self ..." ?

3. This appears to duplicate (improve?) the CSS code we already have. If this adds to the existing code, why is it in a new class instead of with the existing CSS classes?

PR #683, opened 2020-07-20
Remove prefix header from `SVGKit.podspec`

karimhm:
This fixes an issue that was caused by removing the prefix header without propagating the changes to `SVGKit.podspec`


karimhm:
@adamgit @satindar This should fix everything

PR #693, opened 2020-09-09
Fix Demo-OSX build

dtweston:
Even though they look UIKit specific, these header files use macros to also work on macOS


adamgit:
Good point. I feel we should also rename the containing folder from something other thatn "UIKit Additions".

Is there a good common name that captures the platforms where this is relevant?

dtweston:
Maybe it would be better to just call it "ImageViews"?

adamgit:
Sounds good to me. Can you combine with this and I can put it through in one go?

dtweston:
Done.

PR #694, opened 2020-09-09
Revert "Disambiguate Document.h import"

dtweston:
Reverts SVGKit/SVGKit#690

This change breaks the build with an error like:
```
In file included from /Users/dtweston/src/SVGKit/Source/DOM classes/SVG-DOM/SVGTextPositioningElement.m:4:
In file included from /Users/dtweston/src/SVGKit/Source/DOM classes/Unported or Partial DOM/SVGElement_ForParser.h:3:
In file included from /Users/dtweston/src/SVGKit/Source/Parsers/SVGKParseResult.h:8:
/Users/dtweston/src/SVGKit/Source/DOM classes/SVG-DOM/SVGDocument.h:21:1: error: duplicate interface definition for class 'SVGDocument'
@interface SVGDocument : Document
^
In module 'SVGKit' imported from /Users/dtweston/src/SVGKit/Source/DOM classes/SVG-DOM/SVGDocument.h:18:
/Users/dtweston/src/SVGKit/Source/DOM classes/SVG-DOM/SVGDocument.h:21:12: note: previous definition is here
@interface SVGDocument : Document
           ^
```

### DIFF
--- a/SVGKit-iOS.xcodeproj/project.pbxproj
+++ b/SVGKit-iOS.xcodeproj/project.pbxproj
@@ -3668,6 +3668,7 @@
 				OTHER_CFLAGS = "-fembed-bitcode";
 				PRODUCT_NAME = "$(TARGET_NAME).${DYLIB_CURRENT_VERSION}";
 				SKIP_INSTALL = YES;
+				USER_HEADER_SEARCH_PATHS = "$(PROJECT_DIR)/Source";
 			};
 			name = Debug;
 		};
@@ -3691,6 +3692,7 @@
 				);
 				PRODUCT_NAME = "$(TARGET_NAME).${DYLIB_CURRENT_VERSION}";
 				SKIP_INSTALL = YES;
+				USER_HEADER_SEARCH_PATHS = "$(PROJECT_DIR)/Source";
 			};
 			name = Release;
 		};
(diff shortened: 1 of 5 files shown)


adamgit:
If it's just Apple failing to notice that it's already seen that interface during this build run - better perhaps to #ifdef / #def the SVGDocument header? (manual #define guard)?

dtweston:
Adding an include guard does also fix the issue. I opened a new pull request #696 

While this does fix the issue, it doesn't seem ideal. A better fix might be to disable header maps and use the correct paths everywhere to locate the file to import.

dtweston:
> Adding an include guard does also fix the issue. I opened a new pull request #696
> 
> While this does fix the issue, it doesn't seem ideal. A better fix might be to disable header maps and use the correct paths everywhere to locate the file to import.

Whoops, my bad. The include guard **doesn't** fix the issue. I *may* have tested it locally with both the revert and the include guard, which made me think it did.

adamgit:
So, to clarify, current situation is:

1. Disambiguated headers is working for some platforms, but breaking others
2. Where it's broken: include-guards aren't fixing it
3. This PR reverts the disambiguation, but leaves open the possibility of someone attempting disambiguation again in future, if they can demonstrate that their change doesn't re-create these problems

?

If so: for future reference, what platforms are succeeeding with the disambiguation and which are failing? (so we can point any future PRs at this and make sure they get tested against both!)

PR #696, opened 2020-09-10
Add include guards to prevent duplicate definition error

dtweston:
This fixes the duplicate definition error that was introduced by #690


dtweston:
As I mentioned in #694, I think a better fix to disambiguate these header files would be to turn off the header maps that Xcode generates (which allows it to find headers by importing just the filename) and modify all imports to use the correct relative path.

PR #701, opened 2020-10-21
Fix version of CocoaLumber

### LINKED ISSUE #700: SPM dependency resolution fails on target integrity
### New Issue Checklist

* [] I have reproduced the problem while running in Xcode's Debugger, and used breakpoints to examine the actual values of variables vs the expected values
* [x] I am not using CocoaPods, or: I am using CocoaPods, but have verified my podspec points to SVGKit v2.x

When adding SVGKit as an SPM dependency via Package.swift, dependency resolution fails due to CocoaLumberjack supporting iOS 9 while SVGKit supports iOS 8:

```
The package product 'CocoaLumberjack' requires minimum platform version 9.0 for the iOS platform, but this target supports 8.0
```

Tried on XCode 12.0.1.

Package.swift:
```
// swift-tools-version:5.3
// The swift-tools-version declares the minimum version of Swift required to build this package.

import PackageDescription

let package = Package(
    name: "SVGKitTest",
    products: [
        // Products define the executables and libraries a package produces, and make them visible to other packages.
        .library(
            name: "SVGKitTest",
            targets: ["SVGKitTest"]),
    ],
    dependencies: [
        .package(url: "https://github.com/SVGKit/SVGKit", .revision("d1630b5"))
    ],
    targets: [
        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
        // Targets can depend on other targets in this package, and on products in packages this package depends on.
        .target(
            name: "SVGKitTest",
            dependencies: ["SVGKit"]),
        .testTarget(
            name: "SVGKitTestTests",
            dependencies: ["SVGKitTest"]),
    ]
)
```

drademacher:
Minor Version 3.7.0 has breaking changes that lead to problems 
Commit fixes #700

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -19,7 +19,7 @@ let package = Package(
         )
     ],
     dependencies: [
-        .package(url: "https://github.com/CocoaLumberjack/CocoaLumberjack.git", .upToNextMajor(from: "3.0.0"))
+        .package(url: "https://github.com/CocoaLumberjack/CocoaLumberjack.git", .upToNextMinor(from: "3.6.0"))
     ],
     targets: [
         .target(


adamgit:
Why closed?

PR #702, opened 2020-10-23
Update Package.swift

gouen95:
support xcode 12, swift 5.3 and drop support for ios 8

https://github.com/CocoaLumberjack/CocoaLumberjack/releases/tag/3.7.0

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -19,7 +19,7 @@ let package = Package(
         )
     ],
     dependencies: [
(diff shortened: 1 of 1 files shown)


adamgit:
There's only a single change in here - updating the CL version dependency - is that correct?

(dropping support for ios 8 is fine, just checking there's nothing else that was intedned to be included here)

gouen95:
Yes just updating dependencies version target. It would solve including SVGKit in xcode 12 swift 5.3

PR #703, opened 2020-11-13
Fix compilation issues on `3.x` branch

CRoig:
I imported SVGKit as SPM on my project and I found out some compilation issues. This PR fix them:

- Increase Package min version to 9.0 to be used on XCode 12
- Fix broken aliases (on `ImageViews` folder)
- Import UIKit library to define UIColor -> Maybe this breaks MACOS compilation, I am not sure. In any case, MACOS compilation is broken either since UIColor is not valid ¯\\\_(ツ)\_/¯. At least, iOS compilation works now.
- Import `Document.h` properly

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -5,7 +5,7 @@ let package = Package(
     name: "SVGKit",
     platforms: [
         .macOS(.v10_10),
-        .iOS(.v8),
+        .iOS(.v9),
         .tvOS(.v9)
     ],
     products: [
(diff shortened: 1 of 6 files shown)


paulofierro:
This is pretty much [the same changes I had to make](https://github.com/persona-id/SVGKit/commit/b9abd99da6a925f074d2c2d9dfc302b86ce6a411) to bring in SVGKit via SPM too. 

Thank you for submitting a PR! 👏😄

PR #704, opened 2020-11-17
在iOS10.3.1和iOS11.4中闪退,猜测iOS12以下均会如此

1070824493:
会抛出数组越界


adamgit:
It's a good idea, but we should still report these as errors somehow.

(these situations need to inject an error into the SVGKParseResult object - otherwise there is no indication that the SVG file is incorrect)

Issue #715, opened 2021-04-18
Warnings in Swift 5.3 and xCode 12


22Round:
### New Issue Checklist

* [X] I have read and understood the [CONTRIBUTING guide](https://github.com/SVGKit/SVGKit/blob/2.x/CONTRIBUTING.md)
* [X] I have reproduced the problem while running in Xcode, and read the Console log, and acted upon any relevant error or warning messages
* [X] I have reproduced the problem while running in Xcode's Debugger, and used breakpoints to examine the actual values of variables vs the expected values
* [ ] I am not using CocoaPods, or: I am using CocoaPods, but have verified my podspec points to SVGKit v2.x

### Issue Description and Steps
Installed with pod, version 3.0 & getting this warning in swift 5.3.

<img width="383" alt="Screenshot 2021-04-19 at 00 07 31" src="https://user-images.githubusercontent.com/5682469/115159370-de064b80-a0a3-11eb-877f-d2b6b47b8a2d.png">

Is SVGKit still supported project?

Issue #687, opened 2020-08-14
Multiple "Umbrella header for module 'SVGKit' does not include header [...] warnings


adamjansch:
### New Issue Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/SVGKit/SVGKit/blob/2.x/CONTRIBUTING.md)
* [x] I have reproduced the problem while running in Xcode, and read the Console log, and acted upon any relevant error or warning messages
* [x] I have reproduced the problem while running in Xcode's Debugger, and used breakpoints to examine the actual values of variables vs the expected values
* [ ] I am not using CocoaPods, or: I am using CocoaPods, but have verified my podspec points to SVGKit v2.x

### Issue Description and Steps

We have added SVGKit to project via CocoaPods using the following line in our podfile:

    pod 'SVGKit', :git => 'https://github.com/SVGKit/SVGKit.git', :branch => '3.x'

When we build our project Xcode produces multiple compiler errors relating to SVGKit headers:

> Umbrella header for module 'SVGKit' does not include header 'SVGKDefine_Private.h'
> Umbrella header for module 'SVGKit' does not include header 'SVGKImageRep.h'
> Umbrella header for module 'SVGKit' does not include header 'SVGKExporterNSImage.h'

Any ideas how to fix the warnings (without just hiding them)?

adamgit:
Side effect of @karimhm 's recent changes to add Swift Package Mnager support?

karimhm:
@adamgit I confirm that this is related to the recent changes to add Swift Package Manager support.
Happens only on pure 100% Objective-C projects.
 Im working on a fix.

karimhm:
@adamgit I confirm that this is related to the recent changes to add Swift Package Manager support.
This happens only on pure/100% Objective-C projects.
Im working on a fix.

fjtrujy:
Hello,
Just adding some more info.
I'm suffering this issue as well, but I'm using a 100% Swift project, Cocoapods 1.9.3 and Xcode 11.6

Thanks.

timbms:
Yes, here as well. on Xcode 12, almost 100% swift project, Cocoapods 1.10.0.rc.1
We get exactly the same warnings.

adamgit:
If @karimhm can't fix it, I'm not sure what to do here. I looked into what Swift is doing and why SPM causes these problems, but at the time no-one had an easy fix - basically: with the arrival of the SPM it's tough getting Swift and ObjC to work well together.

[...]

PR #720, opened 2021-06-18
Update podspec to make unused header files private

Haoocen:
This PR addresses #687, #715

> Umbrella header for module 'SVGKit' does not include header 'SVGKDefine_Private.h'
Umbrella header for module 'SVGKit' does not include header 'SVGKImageRep.h'
Umbrella header for module 'SVGKit' does not include header 'SVGKExporterNSImage.h'

We are experiencing the same issue since the above three header files are public (Cocoapods default?) while not included in the umbrella header.
And our project treats warnings as errors.

Fix is to declare them as private headers in podspec if not included in `SVGKit.h`

### DIFF
--- a/SVGKit.podspec
+++ b/SVGKit.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name        = 'SVGKit'
-  s.version     = '2.1.0'
+  s.version     = '3.1.0'
   s.license     = 'MIT'
   s.osx.deployment_target = '10.9'
   s.ios.deployment_target = '5.0'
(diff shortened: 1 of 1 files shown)


Haoocen:
Hi @adamgit, since this is merged, do you guys have plan to release this to Cocoapods soon?

adamgit:
Currently no one supports the CocoaPods build - I recommend avoiding CP for
all projects wherever possible (it's a bad tech, bad team, long history of
creaiting nasty bugs).

If you want to take on support for the CP files - podspec etc - please do!

PR #723, opened 2021-06-28
Remove spaces from paths, fix xcodeprojs

dostrander:
**What?**
This purely changes the paths to the source files. The only "code" that has been changed is the xcodeproj's xml to represent the updated paths.

**Why?**

This isn't a problem with SVGKit per se, however with bazel's runfiles they do not currently support spaces and does not seem to be on their roadmap to support it currently (https://github.com/bazelbuild/bazel/issues/4327). This is the only reason we need to have this change in. 

If this is not something you are interested in changing in your repo that's okay and we can likely do a work around Please let me know!


adamgit:
That is WILD. The idea that anyone wouldnt support spaces in paths in this 
century is pretty amazing, let alone 2021!

... I wonder what bazel does when it sees other classic Mac paths, eg 
including colons in filenames (not common but not particularly rare on Mac, 
since it's been supported for decades - I see it a lot with files edited by 
writers and graphic designers).

This late in the project it seems pretty disruptive to make the folder 
structure harder to read just to workaround a bug someone else should 
REALLY fix in their code. Spaces in folders are a very common thing in Mac 
projects, and I strongly suspect that if we do this patch it will only 
break again in near.future when a new contributor adds a new folder, 
unaware of the problem.

So I'm leaning towards rejecting this one. Does anyone else feel strongly 
either way?

PR #729, opened 2021-09-06
fix: can not catch exception about array addObject function

ianhwu:
Array addObject function's exception can not be catched.
when the newChild is nil, will crash at `[self.childNodes.internalArray addObject:newChild];`.
So add a exception for `@catch`.

test url: https://upload.wikimedia.org/wikipedia/de/a/a3/Stoke_City.svg

```
@try
{
libXmlParserParseError = xmlParseChunk(ctx, buff, (int)bytesRead, 0);
}
@catch( NSException* e )
{
    SVGKitLogError( @"Exception while trying to parse SVG file, will store in parse results. Exception = %@", e);
    [currentParseRun addParseErrorFatal:[NSError errorWithDomain:@"SVGK Parsing" code:32523432 userInfo:@{NSLocalizedDescriptionKey:[NSString stringWithFormat:@"Exception = %@", e]}]];
}
```

### DIFF
--- a/Source/DOM classes/Core DOM/Node.m
+++ b/Source/DOM classes/Core DOM/Node.m
@@ -278,6 +278,10 @@
 
 -(Node*) appendChild:(Node*) newChild
 {
+    if (!newChild) {
+        [NSException raise:@"newChild is null" format:@"object cannot be nil"];
+    }
+    
 	[self.childNodes.internalArray removeObject:newChild]; // required by spec
 	[self.childNodes.internalArray addObject:newChild];
 	
--- a/Source/Parsers/SVGKParser.m
+++ b/Source/Parsers/SVGKParser.m
@@ -478,6 +478,8 @@ static void processingInstructionSAX (void * ctx,
 			
 			/** Parser Extenstion creates a node for us */
 			Node* subParserResult = [subParser handleStartElement:name document:source namePrefix:prefix namespaceURI:XMLNSURI attributes:attributeObjects parseResult:self.currentParseRun parentNode:_parentOfCurrentNode];
+            NSAssert( subParserResult != nil, @"Found a tag (prefix:%@ name:%@) that is failing to return a valid node", prefix, name );
+            
 			
 #if DEBUG_XML_PARSER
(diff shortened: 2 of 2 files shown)


ianhwu:
the handleStartElement func may return nil, then call appendChild function will make this crash.
```
		/**
		 Now we know it's a specific parser, check if it handles this particular node
		 */
		if( [[subParser supportedNamespaces] containsObject:XMLNSURI]
		   && [[subParser supportedTags] containsObject:name] )
		{
			[_stackOfParserExtensions addObject:subParser];
			
			/** Parser Extenstion creates a node for us */
			Node* subParserResult = [subParser handleStartElement:name document:source namePrefix:prefix namespaceURI:XMLNSURI attributes:attributeObjects parseResult:self.currentParseRun parentNode:_parentOfCurrentNode];
#if DEBUG_XML_PARSER
			SVGKitLogVerbose(@"[%@] tag: <%@:%@> id=%@ -- handled by subParser: %@", [self class], prefix, name, ([((Attr*)[attributeObjects objectForKey:@"id"]) value] != nil?[((Attr*)[attributeObjects objectForKey:@"id"]) value]:@"(none)"), subParser );
#endif
			
			/** Add the new (partially parsed) node to the parent node in tree
			 
			 (need this for some of the parsing, later on, where we need to be able to read up
			 the tree to make decisions about the data - this is REQUIRED by the SVG Spec)
			 */
			[_parentOfCurrentNode appendChild:subParserResult]; // this is a DOM method: should NOT have side-effects
			_parentOfCurrentNode = subParserResult;
			
			if( parsingRootTag )
			{
				currentParseRun.parsedDocument.rootElement = (SVGSVGElement*) subParserResult;
                [self addExternalStylesheetsToRootElement];
			}
			
			return;
		}
```

adamgit:
The comments look good but they dont appear related to the code change. Did tou submit the wrong commit by accident?

In your “handleStartElement” example (good spot!) I think we should be doing an NSAssert that subParserResult is not-nil. If a subparser is failing to return a valid node then something really bad has happened.

ianhwu:
hi, i add the assert, but if app is released, [array addObject: nil]; will crash, and this exception can not  be catch by below.
```
@try
{
libXmlParserParseError = xmlParseChunk(ctx, buff, (int)bytesRead, 0);
}
@catch( NSException* e )
{
    SVGKitLogError( @"Exception while trying to parse SVG file, will store in parse results. Exception = %@", e);
    [currentParseRun addParseErrorFatal:[NSError errorWithDomain:@"SVGK Parsing" code:32523432 userInfo:@{NSLocalizedDescriptionKey:[NSString stringWithFormat:@"Exception = %@", e]}]];
}
```
So, i have to throw a exception to avoid this.

PR #747, opened 2022-04-07
fix: basic contentMode support for SVGKFastImageView

farfromrefug:
This PR adds support for `UIViewContentModeScaleAspectFill` and `UIViewContentModeScaleAspectFit`


farfromrefug:
@calbot you need to give more.info.

PR #764, opened 2022-10-18
remove cocoalumberjack

sisoje:


adamgit:
This appears to simply remove all logging, and provide no replacement? (have I misread or misunderstood the suggested change here?)

If that's correct - are you volunteering to take on all incoming support requests? Every single person who makes the smallest mistake in their code, or has any problem -- can we give them your private email address for you to debug their code for them? You're removing their ability to solve their own problems.

bstillitano:
@adamgit an attitude like that will kill your library. Perhaps you could've attempted to reason with @sisoje as to why the change is not appropriate and subsequently he could've made adjustments and found a suitable alternative.

As it stands, your library does not compile and his branch fixes that.

bubudrc:
maybe, we can change the external library of cocolumberjack and use the apple support of [OSLog](https://developer.apple.com/documentation/os/oslog)

Issue #772, opened 2022-11-15
Crash when the SVG width/height is set to auto


arturdev:
### New Issue Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/SVGKit/SVGKit/blob/2.x/CONTRIBUTING.md)
* [x] I have reproduced the problem while running in Xcode, and read the Console log, and acted upon any relevant error or warning messages
* [x] I have reproduced the problem while running in Xcode's Debugger, and used breakpoints to examine the actual values of variables vs the expected values
* [ ] I am not using CocoaPods, or: I am using CocoaPods, but have verified my podspec points to SVGKit v2.x

### Issue Description and Steps

In some cases when the SVG's width/height is set to `auto` the app crashes: [example svg](https://openseauserdata.com/files/98ab105292ecec6f5524fd8ddc1c3b8e.svg) 

*** Terminating app due to uncaught exception 'CALayerInvalidGeometry', reason: 'CALayer position contains NaN: [nan nan]'
I've found that if in `CSSPrimitiveValue.m` in the case of the unhandled `CSS_STRING` I return a non-zero value (f.e. 600) the app works correctly and SVG is rendered correctly.

PR #773, opened 2022-11-15
#772 fixed crash when SVG width/height is set to "auto"

arturdev:


adamgit:
The spec seems to say that 'auto' is an illegal value, and that if in doubt, we should be substituting '100%':

"If the attribute is not specified, the effect is as if a value of '100%' were specified."

?

https://www.w3.org/TR/SVG11/struct.html#SVGElementWidthAttribute

adamgit:
(I vaguely remember 'auto' being valid, it's just that a quick skim of the attribs suggest not. Setting it to 100% seems sensible as the fallback though - and seems to be what e.g. chrome is doing when opening the example.svg you provided in the issue report. Can we just get this code to act as 100%?)

arturdev:
Well, this is just a quick workaround for the crash rather than an actual fix till someone who has more knowledge of SVGKit fixes it in the right way :)

Issue #775, opened 2022-12-08
`CocoaLumberjack` build issues


lxalfonso:
### New Issue Checklist

* [✅] I have read and understood the [CONTRIBUTING guide](https://github.com/SVGKit/SVGKit/blob/2.x/CONTRIBUTING.md)
* [✅] I have reproduced the problem while running in Xcode, and read the Console log, and acted upon any relevant error or warning messages
* [N/A] I have reproduced the problem while running in Xcode's Debugger, and used breakpoints to examine the actual values of variables vs the expected values
* [ ✅] I am not using CocoaPods, or: I am using CocoaPods, but have verified my podspec points to SVGKit v2.x

### Issue Description and Steps

Hey everyone! I was wondering if anyone's still had the `CocoaLumberjack` minimum version issue that has been brought up here before. My project's using SPM and currently I get either one of two errors:
A. If my dependency is set on **Version: Up to Next Major `3.0.0`**, then I see: 
```
The package product 'CocoaLumberjack' requires minimum platform version 11.0 for the iOS platform, but this target supports 9.0
```
B. If my dependency is set on the branch `3.x`, then I see:
```
'CocoaLumberjack/CocoaLumberjack.h' file not found
```

Not sure if I'm missing something, but has anyone encountered anything similar?

bstillitano:
Same issue for me on branch `3.x`

bubudrc:
same issue here..

airspeed:
Be sure to import `CocoaLumberjackSwift` as well:
https://github.com/CocoaLumberjack/CocoaLumberjack#swift-package-manager

bubudrc:
but... if you set, on the Package file the dependency, this should be automatically added, and that it's what happened. The error is other

lxalfonso:
> Be sure to import `CocoaLumberjackSwift` as well:
> https://github.com/CocoaLumberjack/CocoaLumberjack#swift-package-manager

@airspeed Are you referring to something that we would be able to do inside of our own projects? The error I'm referring to is coming from the SVGKit source code, and in an Objective-C file at that. I'm not sure if `CocoaLumberjackSwift` applies here.

airspeed:
@bubudrc @lxalfonso I needed to extra import `CocoaLumberjack` as Package in *my* project, and from the list of modules I selected both `CocoaLumberjack` and `CocoaLumberjackSwift`.
This *in addition* to `SVGKit` importing `CocoaLumberjack`. Now my projects builds.

[...]

PR #780, opened 2023-02-07
CocoaLumberjack Import Issue Fix

lxalfonso:
As described in issue #775 there was an issue making new builds with SVGKit on the `3.x` branch because of an error importing `CocoaLumberjack` via a pre-compiler import. After doing some testing, it looks like switching to using an explicit module import resolves the issue.

### DIFF
--- a/Source/SVGKDefine_Private.h
+++ b/Source/SVGKDefine_Private.h
@@ -8,7 +8,7 @@ SVGKDefine define some common macro used for private header.
 #define SVGKDefine_Private_h
 
 #import "SVGKDefine.h"
-#import <CocoaLumberjack/CocoaLumberjack.h>
+@import CocoaLumberjack;
 
 // These macro is only used inside framework project, does not expose to public header and effect user's define
 


adamgit:
Didn't this cause a problem with Swift builds? I thought we tried this and it didn't work (broke SPM?). Can you confirm you've tested with Swift and the SPM, and it's working OK?

lxalfonso:
> Didn't this cause a problem with Swift builds? I thought we tried this and it didn't work (broke SPM?). Can you confirm you've tested with Swift and the SPM, and it's working OK?

@adamgit So far it seems to work with Swift builds perfectly. I had originally tested with the OSX-demo app which worked (after some modification since it seemed to be having an issue resolving some references to `contentMode` but I didn't have a lot of time to look into it) and then I loaded it as a local package into another test iOS app which also worked perfectly.

But to be sure I ran the same test project with my fork of the package added via SPM (on the branch from this PR, `issues/775`) and it also seems to work great. I'll attach a screenshot the app running.  If there's any other edge case scenarios that would be cause for concern please let me know, I want to make sure this fixes the issue at hand not introduce anymore issues.

Screen Shot 2023-02-08 at 9 58 44 AM

adamgit:
Awesome, thanks.

I'll merge this now - let's keep an eye on the Issues, and see if anyone reports problems, but hopefuly this is finally solved!

NiliStein:
Hi!
Sorry to be the one raising this issue from the dead 🥲 
Actually this commit causes our mixed (swift + objc) project to not compile. More in detail, I get the following errors for the following lines in the **CocoaLumberjack header** and a few more DD__ headers as well, all claimed to be under the SVGKit framework:
```
#import <CocoaLumberjack/DDLegacyMacros.h>
```
```
Include of non-modular header inside framework module 'CocoaLumberjack.CocoaLumberjack': '/[project path]/Pods/Headers/Public/CocoaLumberjack/DDLegacyMacros.h'
```
I have around 8 error of the same type each for a different import.

I have currently set the version of the podfile to the previous working commit, but I would appreciate help in solving this one to enjoy the newer versions as well :)

cikichen:
Module 'CocoaLumberjack' not found
```
/**
SVGKDefine_Private.h

SVGKDefine define some common macro used for private header.
*/

#ifndef SVGKDefine_Private_h
#define SVGKDefine_Private_h

#import "SVGKDefine.h"
@import CocoaLumberjack;

// These macro is only used inside framework project, does not expose to public header and effect user's define

#define SVGKIT_LOG_CONTEXT 556

#define SVGKitLogError(frmt, ...)   LOG_MAYBE(NO,                LOG_LEVEL_DEF, DDLogFlagError,   SVGKIT_LOG_CONTEXT, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
#define SVGKitLogWarn(frmt, ...)    LOG_MAYBE(LOG_ASYNC_ENABLED, LOG_LEVEL_DEF, DDLogFlagWarning, SVGKIT_LOG_CONTEXT, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
#define SVGKitLogInfo(frmt, ...)    LOG_MAYBE(LOG_ASYNC_ENABLED, LOG_LEVEL_DEF, DDLogFlagInfo,    SVGKIT_LOG_CONTEXT, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
#define SVGKitLogDebug(frmt, ...)   LOG_MAYBE(LOG_ASYNC_ENABLED, LOG_LEVEL_DEF, DDLogFlagDebug,   SVGKIT_LOG_CONTEXT, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
#define SVGKitLogVerbose(frmt, ...) LOG_MAYBE(LOG_ASYNC_ENABLED, LOG_LEVEL_DEF, DDLogFlagVerbose, SVGKIT_LOG_CONTEXT, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)

#if DEBUG
static const int ddLogLevel = DDLogLevelVerbose;
#else
static const int ddLogLevel = DDLogLevelWarning;
#endif

#if SVGKIT_MAC
#define NSStringFromCGRect(rect) NSStringFromRect(rect)
#define NSStringFromCGSize(size) NSStringFromSize(size)
#define NSStringFromCGPoint(point) NSStringFromPoint(point)
#endif

#endif /* SVGKDefine_Private_h */
```

Issue #793, opened 2023-07-18
Crash when SVG url doesn't return an image


JoeFerrucci:
### New Issue Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/SVGKit/SVGKit/blob/2.x/CONTRIBUTING.md)
* [x] I have reproduced the problem while running in Xcode, and read the Console log, and acted upon any relevant error or warning messages
* [x] I have reproduced the problem while running in Xcode's Debugger, and used breakpoints to examine the actual values of variables vs the expected values
* [x] I am not using CocoaPods, or: I am using CocoaPods, but have verified my podspec points to SVGKit v2.x

### Issue Description and Steps

I'm getting a crash when trying this url: https://upload.wikimedia.org/wikipedia/commons/7/7c/Pistons_logo17.svg

`let svgImage = SVGKImage(contentsOf: svgUrl)`

Here's the crash log. I don't have a solution for this:
![Screenshot 2023-07-18 at 6 16 40 PM](https://github.com/SVGKit/SVGKit/assets/2114494/a81ed2a2-b736-45df-94c5-07d87073d4e1)

Here is the result when opening the URL in a web browser:

![Screenshot 2023-07-18 at 6 18 44 PM](https://github.com/SVGKit/SVGKit/assets/2114494/5e8d7d7f-41e3-4f60-9539-f9954780388b)

Here is Xcode console upon crash:

```bash
libc++abi: terminating due to uncaught exception of type NSException
Message from debugger: Terminated due to signal 6
```

Error Log:

```bash
Error internally in Apple's NSData trying to read from URL 'https://upload.wikimedia.org/wikipedia/commons/7/7c/Pistons_logo17.svg'. Error = Error Domain=NSCocoaErrorDomain Code=256 "The file “Pistons_logo17.svg” couldn’t be opened." UserInfo={NSURL=https://upload.wikimedia.org/wikipedia/commons/7/7c/Pistons_logo17.svg}
```

Assertion:

```bash
Attempted to init an SVGKImage using a nil SVGKSource
```

JoeFerrucci:
I see in your code there's a bunch of logic to check for nil streams / errors, but unfortunately it still crashes my app. I wish it would fail gracefully instead of crashing -- just return a nil SVGImage instead of crashing.

JoeFerrucci:
For anyone who stumbles upon this before it's fixed in SVGKit:

Construct the SVGKSource yourself, check for nil.

[...]

PR #794, opened 2023-07-18
Log error instead of crashing - Fix Issue #793

JoeFerrucci:
I've made it so it won't crash (Assert) when the source is nil. It will just log the error using SVGKit's internal error logger.

Reference filed issue for any details and follow-up dialog: https://github.com/SVGKit/SVGKit/issues/793

I've tested this and it's no longer crashing. To see the error, `SVGKit.enableLogging()`

I'm was getting a crash when trying this url: https://upload.wikimedia.org/wikipedia/commons/7/7c/Pistons_logo17.svg

`let svgImage = SVGKImage(contentsOf: svgUrl)`

Here's the crash log:
Screenshot 2023-07-18 at 6 16 40 PM

Here is the result when opening the URL in a web browser:

Screenshot 2023-07-18 at 6 18 44 PM

Here is the Xcode console upon crash:

```bash
libc++abi: terminating due to uncaught exception of type NSException
Message from debugger: Terminated due to signal 6
```

Error Log:

```bash
Error internally in Apple's NSData trying to read from URL 'https://upload.wikimedia.org/wikipedia/commons/7/7c/Pistons_logo17.svg'. Error = Error Domain=NSCocoaErrorDomain Code=256 "The file “Pistons_logo17.svg” couldn’t be opened." UserInfo={NSURL=https://upload.wikimedia.org/wikipedia/commons/7/7c/Pistons_logo17.svg}
```

Assertion:

```bash
Attempted to init an SVGKImage using a nil SVGKSource
```

### DIFF
--- a/Source/SVGKImage.m
+++ b/Source/SVGKImage.m
@@ -330,8 +330,12 @@ static NSMutableDictionary* globalSVGKImageCache;
 }
 
 - (id)initWithSource:(SVGKSource *)newSource {
-	NSAssert( newSource != nil, @"Attempted to init an SVGKImage using a nil SVGKSource");
-	
+    if( newSource == nil )
+    {
+        SVGKitLogError(@"Attempted to init an SVGKImage using a nil SVGKSource");
+        return nil;
+    }
+
 	self = [self initWithParsedSVG:[SVGKParser parseSourceUsingDefaultSVGKParser:newSource] fromSource:newSource];
 	
 	return self;


adamgit:
You should never remove an Assert from code.

Your issue appears to be that something is creating a nil SVGKSource (which is an illegal action) - the fix would be to find the code that returns the nill source, and have it return a non-nil source while logging an error.

adamgit:
Also: assertions are NOT crashes. They only happen in 'developer' builds, and they're there to tell you that something VERY wrong has happened, and that you shouldn't continue with your development without resolving it.

App Store builds won't 'crash', unless you actively force them to (by enabling assertions in your release build, which you shouldn't do).

If something is broken with SVGKit making builds that include assertions in Release (!) that would be a much bigger problem / urgent issue that needs fixing, because it affects hundreds of Asserts in the codebase

JoeFerrucci:
I understand the point of an assertion and that an assertion most definitely leads to a crash. 

If you provide a convenience init from a URL and can’t handle the URL being bad, or provide a way to get feedback…that’s bad design. Sorry.

I disagree. Thanks for considering the fix. In my Issue 793 I provide the workaround to the problem (yes it’s a problem).

JoeFerrucci:
You’re right. Thanks.

adamgit:
Yeah, the SVGKSource that is handling the URL should be a lot better.

From memory: it was originally just one line of code, and it got a *little*
improved over time, but it's always been a "someone should take the time to
design/implement URL handling properly, and flesh out this class" situation.

PR #797, opened 2023-08-19
Loading

duldul09:


adamgit:
There are too many undocumented + unrelated changes in this PR, including a lot that are merely whitespace/newline changes. I cannot review it, it's too much for one commit.

Please re-submit in smaller chunks, where it's clear what you're changing/fixing in each one. Especially: put all the non-siginificant (whitespace, newline, etc) changes in a single commit so we can focus on the meaningful code changes.

From reading the first 20% it looks like there's a total of maybe 5-8 separate feature/fix changes in here, so I'd expect approximately that many commits (or slightly less if it's obvious that a few changed all relate to the same problem.

Thanks!

Issue #688, opened 2020-08-14
Removing CocoaLumberjack as a dependency


freak4pc:
### Issue Description and Steps

While this was discussed previous in issue such as #573, I'd like to reiterate this request.

Forcing/requiring a specific logging framework as a third-party dependency for a project is really ill-advised. If I'm not a consumer of CocoaLumberjack, you've just added a huge dependency to my codebase for only wanting to draw SVGs on screen. 

I would highly suggest removing CocoaLumberjack as a direct dependency and think of one of two options:

## Option 1: Abstracton

1. Make the logging functionality a lightweight protocol, only notifying that the framework wishes to log an event
2. Allow consumers to conform to the protocol if they want to get the logging events, and let the consumers choose how they wish to actually perform the logging. The lightweight protocol can default to something such as printing or `os_log`ing (on supported platforms) 

## Option 2: Apple Unified Logging

Use `OSLog` where applicable instead of the other options. The downside here is that this project supports very old OS versions, which os_log doesn't support, so it would require some more thought. 

I'd appreciate your thoughts on this, and hope you can understand why forcing such a fundamental dependency (logging) for a consumer of a drawing library seems quite aggressive.

Thank you!

adamgit:
In 2020 I think it's fair to switch to OSLog.

But we should do a tagged release first, so that anyone dependent on it can
pin to that version.

(abstraction would be best, but ... it takes some effort to design (make
sure the interfaces supports all major logging frameworks reasonably well)
+ implement (change all the code throughout, test it all!). If someone
volunteered to do it, great.)

However, I'll also note that Lumberjack has always been a lightweight
logging system - it's one of the reasons we chose it 7+ years ago, it was
fast (much faster than iOS's built-in logging!) and worked well on low-end
devices (I think we were still using iPhone4's and iPad1's back then). So
the reason it's not been abstracted away already is ... it just hasn't
impacted performance and/or binary size enough for anyone to care enough to
make the change.

freak4pc:
Thanks Adam, appreciate your quick response
If you're OK with not having logging before iOS 10, that would probably be the "cheapest" way to remove CLJ.

I don't have any issues with CLJ in itself, but logging is Infra, and in most apps' perspectives - Infra should be the decision of the app and not third-party frameworks

In your perspective, when would be a good time to cut a point release so this change can be made?

ray007:
[...]

PR #799, opened 2023-08-30
Replace CocoaLumberjack with OSLog

squarefrog:
This PR replaces CocoaLumberjack with OSLog, as requested in https://github.com/SVGKit/SVGKit/issues/688. 

Note that you can set the minimum log level in Xcode from the command-line. [Here's a guide](https://stackoverflow.com/a/45988511). In Xcode 15 this is even easier, as you can filter the logs by category, subsystem or level direct from the console. I recommend watching [Debug with structured logging](https://developer.apple.com/videos/play/wwdc2023/10226/) from WWDC '23 for more details.

Here's a guide to the levels, as they don't quite match up with `DDLog`: https://notificare.com/blog/2020/06/12/Using-Unified-Logging-in-iOS-or-tvOS-apps/

This is not ready for merge, as there are a couple of things that need a cleanup. 

1. When building in Xcode 14, I get around 300+ warnings concerning the use of quotes instead of angle brackets for imports. This can be rectified by setting `CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER` to false.
2. I haven't written Objective-C in almost a decade, so I'm unsure how we can make the logger a static instance rather than creating each time.
3. `[SVGKit enableLogging];` needs cleaning up

I've split the changes into commits to make it easier to review, but it's still a big one.


Source/SVGKDefine_Private.h
```diff
@@ -8,23 +8,18 @@ SVGKDefine define some common macro used for private header.
...
 // These macro is only used inside framework project, does not expose to public header and effect user's define
 
 #define SVGKIT_LOG_CONTEXT 556
 
-#define SVGKitLogError(frmt, ...)   LOG_MAYBE(NO,                LOG_LEVEL_DEF, DDLogFlagError,   SVGKIT_LOG_CONTEXT, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
-#define SVGKitLogWarn(frmt, ...)    LOG_MAYBE(LOG_ASYNC_ENABLED, LOG_LEVEL_DEF, DDLogFlagWarning, SVGKIT_LOG_CONTEXT, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
-#define SVGKitLogInfo(frmt, ...)    LOG_MAYBE(LOG_ASYNC_ENABLED, LOG_LEVEL_DEF, DDLogFlagInfo,    SVGKIT_LOG_CONTEXT, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
-#define SVGKitLogDebug(frmt, ...)   LOG_MAYBE(LOG_ASYNC_ENABLED, LOG_LEVEL_DEF, DDLogFlagDebug,   SVGKIT_LOG_CONTEXT, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
-#define SVGKitLogVerbose(frmt, ...) LOG_MAYBE(LOG_ASYNC_ENABLED, LOG_LEVEL_DEF, DDLogFlagVerbose, SVGKIT_LOG_CONTEXT, nil, __PRETTY_FUNCTION__, frmt, ##__VA_ARGS__)
+#define SVGKitLogger os_log_create("com.github.svgkit", "SVGKit")
```
squarefrog, line 17: This should be a static instance instead, but I'm not sure how to do this with Objective-C macros.

PR #805, opened 2023-12-07
Adding PPIs for new iPads

lihter:


adamgit:
Thanks, but please resubmit with only the affected lines changed. Don't
reformat the project to your custom setup - it generates thousand of lines
of fake changes for everyone else! :)

PR #807, opened 2024-01-04
Fix compiler warning in #687 with Swift Package Manager

### LINKED ISSUE #687: Multiple "Umbrella header for module 'SVGKit' does not include header [...] warnings

Simon-Zeng:
Reorganized private headers and updated SPM package file to fix #687 .


adamgit:
For the logger change, we should really be doing this, I think:

if @available(iOS 10.0, ...) {
  DDLog.add(DDOSLogger.sharedInstance)
} else {
  DDLog.add(DDTTYLogger.sharedInstance)
  DDLog.add(DDASLLogger.sharedInstance)
}

... i.e. don't use DDOS with DDTTY. But ... I'd be happy to approve a PR that removes TTY + ASL entirely - Apple doesn't support anything less than about iOS 16 any more.

timbms:
I was hopeful that those annoying compiler warnings would be gone, but I still get them:
Screenshot 2024-01-10 at 14 17 50
Yes, I am positive that I got the latest 3.x with revision "ba900cc21152bee80c9a86696decf91f1c415cc8"

Simon-Zeng:
@timbms You are right, there are some warnings left still. It is fixed in another [pull request](https://github.com/SVGKit/SVGKit/pull/810)

Issue #802, opened 2023-09-28
Crash on some svg file due to OOM issue


Simon-Zeng:
### New Issue Checklist

* [x] I have read and understood the [CONTRIBUTING guide](https://github.com/SVGKit/SVGKit/blob/2.x/CONTRIBUTING.md)
* [x] I have reproduced the problem while running in Xcode, and read the Console log, and acted upon any relevant error or warning messages
* [x] I have reproduced the problem while running in Xcode's Debugger, and used breakpoints to examine the actual values of variables vs the expected values
* [x] I am not using CocoaPods, or: I am using CocoaPods, but have verified my podspec points to SVGKit v2.x

### Issue Description and Steps

App will consume lots of memory and then crash when viewing some svg files with large viewBox:
`
   version="1.1"
   viewBox="0 0 10000 8500"
   xml:space="preserve"
   id="svg5418"
   inkscape:version="0.48.4 r9939"
   width="100%"
   height="100%"
`
i.e. my svg file is 30KB, but when viewing in app, it cost 2.9 GB memory.
Is there any way to avoid this issue?

adamgit:
If your image was a single white rectangle, at that resolution you have
forced your machine to dedicate at least 255 MB just to send it to the
screen for rendering. Operating Systems often cache rendering, so you might
expect that to be 500+MB instead.

If you're viewing the image as bitmap: Reduce the size of the viewBox. You
are currently *forcing* your machine to spend this much memory - that is
what the viewBox parameter does.

If you're viewing the image as vector: your image probably has too much
complexity (too many gradients, etc) for your machine to handle. You can
reduce the viewbox to make it easier for your OS -- the OS will only have
to generate a smaller number of pixels during rendering. A vector image can
be thousands of times smaller on disk than in-memory.

PR #808, opened 2024-01-04
Fixes crash issues on iOS 17

Simon-Zeng:
In this commit, i had fixed 2 crash issues due to OOM  (#772 , #802 ) when viewing some svg files.


adamgit:
This is a nice idea but 1000x1000 is much much too small - we were displaying images 4000 wide even back on iPhone 5. The issue with #802 is usually that someone tries to convert to bitmap when their app should be loading as Vector (which is a mistake in app-design).

If an app is getting OOM because the SVG is "too large" then that's really a bug in the app - the app should be pre-downscaling the SVG before loading it (either by rendering into a fixed size area that's smaller / using a different Renderer that has fixed memory characteristics, or by changing the SVG viewBox before passing it to SVGKit).

If you're going to override the viewBox that has to be done in the API call where the SVGKImage is loaded - we have something for this already that's used for the zooming in the demo apps, from memory it takes an argument that pre-scales the viewBox during loading/rendering (the change here would be an extra option to 'limit' the viewbox to be e.g. 1000x1000 as in your PR - but applied by the user, explicitly, during SVKImage load).

Simon-Zeng:
You are right, @adamgit. I have reverted changes for viewPort changes. 
Another crash issue is that app may crash on iOS 17 when using deprecated UIGraphicsBeginImageContextWithOptions.
It would be nice to have fixes for it.

PR #822, opened 2024-03-13
Fix imports

g4rb4g3:
We use SVGKit in a react-native project that comes with Yoga which brings its own Node.h that SVGKit tries to make use of then. This makes builds fail obviously. 😢  I also added[ another import fix for CocoaLumberjack](https://github.com/troZee/SVGKit/commit/495ead905959a44a386bf06a0525f4aeb3a59722) that @troZee did. :)

Would be great if this could get merged here so we can switch back to the official repo. 🚀  Thanks!

### DIFF
--- a/Source/DOM classes/Core DOM/Attr.h
+++ b/Source/DOM classes/Core DOM/Attr.h
@@ -16,7 +16,7 @@
 #import <Foundation/Foundation.h>
 
 /** objc won't allow this: @class Node;*/
-#import "Node.h"
+#import <Node.h>
 @class Element;
 
 @interface Attr : Node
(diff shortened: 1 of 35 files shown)


adamgit:
Changing it to "Dom*" achieves nothing - there are other libraries that use that name too.

This patch would have to change everything to an SVGKit-specific prefix. Otherwise you're not solving the problem (and note: this change breaks all existing code that references anything in the DOM, for all existing apps: it's a massive change).

We've discussed this in the past, and the ideal is to instead move the DOM code into a separate library. At the time, there were NO open-source libraries providing good-quality DOM on Apple platforms (maybe there are now?).

Until someone is willing to do that (and test it thoroughly), the minimum change is rename everything with an SVGKit prefix (to prevent name clashes in future), but then it will have to be renamed back again if/when someone migrates this to use external DOM library(ies).

(Unless Apple has finally added support for namespaces to ObjC? Which would make everyone's lives much, much, easier!)

PR #826, opened 2024-04-30
Update SVGLength.m

mylogon341:
Correct apparent iPad13,18 typo


mylogon341:
This will fix an issue where a specific iPad is not caught, but it might conflict with #805, (which looks to be dead, however)

PR #833, opened 2024-06-25
chore: Update `Node.h` path into SVGKit/Node.h to fix namespace conflict with React Native Yoga's Node header file

jbagaresgaray:
Update `Node.h` path into SVGKit/Node.h to fix namespace conflict with React Native Yoga's Node header file


ronickg:
Can confirm this fixes the issue for React Native

Issue #834, opened 2024-06-28
Simulator crash `SVGLength.pixelsPerInchForCurrentDevice` calculation


JoeFerrucci:
<img width="1486" alt="Screenshot 2024-06-28 at 8 53 37 AM" src="https://github.com/SVGKit/SVGKit/assets/2114494/a4f2294f-4a2d-4c1f-b59a-6c5719c69ff4">

Looks like it should've been caught on line 234 above where it actually fell.

This is on Xcode 15.4 Simulator running on M2 Max macOS Sonoma 14.5
Its been happening on old versions of Xcode and macOS also. Same machine.

```objc
if( [platform hasPrefix:@"x86_64"])
{
  SVGKitLogWarn(@"[%@] WARNING: you are running on the simulator; it's impossible for us to calculate centimeter/millimeter/inches units correctly", [self class]);
  return 132.0f; // Simulator, running on desktop machine
}
```

The fix is probably changing `if( [platform hasPrefix:@"x86_64"])` to be more complete and efficient at detecting Simulator.

JoeFerrucci:
As a temporary fix I've changed that line to be:

```objc
#if TARGET_IPHONE_SIMULATOR
  SVGKitLogWarn(@"[%@] WARNING: you are running on the simulator; it's impossible for us to calculate centimeter/millimeter/inches units correctly", [self class]);
  return 132.0f; // Simulator, running on desktop machine
#endif
```

Simulator no longer crashes

JoeFerrucci:
I see in the latest `3.x` branch that you've changed it to `if( [platform hasPrefix:@"x86_64"] || [platform hasPrefix:@"arm64"] )`

I have 3.0.0 but I don't have **_that_** code locally.

I guess someone's already fixed this particular issue but failed to provide a proper release; I will use 3.x as directed in the README.

adamgit:
We do formal releases very rarely, usually only when we're about to make
major changes.

(mainly because it's a lot of hassle :))

Igorsnaki:
@adamgit  hello. do you plan to add support  for new devices: iPhone 16/ 16 Pro models. 
There is the same crash on new devices. ( There are no such devices in modelsPPIManifest() method of SVGLength.m )

adamgit:
I don't have those devices. Usually soeone who does submits a pull request.
It takes seconds to do.

darkfiredarkhalo:
[...]

PR #835, opened 2024-06-28
Detect Simulator using pragma. Fixes/improves Issue #834

JoeFerrucci:
Use TARGET_IPHONE_SIMULATOR instead of checking for string prefix. 

Fixes/improves https://github.com/SVGKit/SVGKit/issues/834

### DIFF
--- a/Source/DOM classes/SVG-DOM/SVGLength.m
+++ b/Source/DOM classes/SVG-DOM/SVGLength.m
@@ -594,11 +594,10 @@ static float cachedDevicePixelsPerInch;
         return 326.0f;
     }
     
-    if( [platform hasPrefix:@"x86_64"] || [platform hasPrefix:@"arm64"])
-    {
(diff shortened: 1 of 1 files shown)


adamgit:
I'm sure we originally had this, and there was some reason it didn't work ... maybe it was a bug in Xcode that is now fixed? ... If I remember correctly: Xcode would mark this as true "always" when you tried to auto-build a merged binary .so file (which our custom build script does).

JoeFerrucci:
> I'm sure we originally had this, and there was some reason it didn't work ... maybe it was a bug in Xcode that is now fixed? ... If I remember correctly: Xcode would mark this as true "always" when you tried to auto-build a merged binary .so file (which our custom build script does).

I’ll test it in Release and on device and if it returns true always, I’ll close this.

adamgit:
Testing the #define on its own won't do much - you'd need to test it
while running the build-script, because xcode runs the build-script
with different env-variables compared to just running a basic build
directly.

(Assuming the build script is still in-use! I think it's still there,
making dual-binaries that work in both simulator and in store - but I
haven't looked at it in years, so maybe it's been
deleted/removed/replaced by now!)

PR #879, opened 2026-09-21
Fix SwiftPM CocoaLumberjack version range to 3.7.0–3.9.1

kliuchev:
CocoaLumberjack requires iOS 15+.

The current SwiftPM requirement, `.upToNextMajor(from: "3.7.0")`, allows CocoaLumberjack releases newer than 3.9.1, up to but excluding 4.0.0. This change caps the dependency at 3.9.1 while retaining support for versions starting at 3.7.0. 

```swift
.package(url: "https://github.com/CocoaLumberjack/CocoaLumberjack.git", "3.7.0"..."3.9.1")
```

The closed range includes both endpoints. The change is limited to the dependency requirement in `Package.swift` and follows the existing formatting.

Validation:
- `git diff --check` passes for the PR diff.


sashamyshkina:
👍👍👍👍

